Issue #10, opened 2026-08-19
Plan — what happens in what order, and why the order matters


drake69:
This is the single place where the plan lives, kept edited in place rather than re-explained in comments. Until now it existed in fragments across three threads, and none of you could have reconstructed it — that's on me, and this fixes it.

**Everything in section 2 is a proposal, subject to the answer I asked for in #2 by 31 August.** Section 1 is technical and doesn't depend on that answer. If you want to change something here, argue with the constraint written under it — every step is positioned by a specific reason, and if the reason is wrong the step moves.

## 1. Settled — the technical base

**The base is `omoda9-ha` @ v1.13.0.** It's the non-fork (which HACS's default store requires of the canonical repo), its history is unbroken, and it carries the stars, releases and redirect that keep existing installs working through a rename.

**@GurliGebis's merged repo is the completeness checklist, not the base.** It's a genuine union of the two lines, but it rebuilds ~119 real commits as 8 synthetic ones — which costs `git blame` (Rino's and Jack's attribution disappears) and `git bisect`, our main tool on the four open bugs — and it branches from v1.5.22, so it doesn't contain v1.10–v1.13. It stays private and frozen. One caveat found while extracting the checklist from it: it's missing the `At Home` binary sensor while the energy card it carries references that entity three times. So: a list of things to check, not a line-by-line oracle.

**The fork-only delta to port: 20 files plus 6 features that live inside shared files** — speaking OTP errors, home/away charging energy, the two Lovelace cards, the sandbox and CI work, and the rest. Cherry-pick, not rebase: the two lines diverged at v1.5.22. This one is mine and it's queued; nobody else needs to carry it over.

**Architecture: the axis of variation is the platform, not the brand or the model.** The real boundary is the authentication and signing front door (legend gateway vs the Chery BFF+UAA stack vs CarLinko); regions turn out to be a hostname swap over data; and what a given car accepts is discovered at runtime and changes over time. Written up in #6 as a domain model, a target layout and a route in five independently releasable slices, each locked by a CI assertion so it can't drift back. If you only read one thing here, read that PR instead of this issue.

## 2. The sequence, and why each step sits where it does

**1 — Port the fork-only delta onto the canonical.** *Why first:* every later step gets more expensive with two diverging codebases. When it's done we diff the result against the merged repo's `main`: nothing significant left over means we forgot nothing.

[...]

Issue #8, opened 2026-08-19
CONTRIBUTORS.md — name everyone who built a line of this


drake69:
Everyone who built a line of this gets named, in the canonical repo's first commit. This holds whatever the answer to #2 turns out to be — it isn't a reward for agreeing with me.

I've deliberately not written your entries. I'd get them wrong, and it isn't my place: reply with your own line and I'll paste it in.

Something like:

> **@handle** — what you did, in your own words. Car, if you want it known.

Two things worth saying out loud:

**Field testers count.** Someone who runs a release on their car and reports what happened contributes something no one else can produce. Same file, same standing.

**None of us wrote this by hand.** Whether we say that in the file, and how, is a group decision rather than a default — nobody's tools get credited here without the group agreeing on it. Tell me what you think; I don't have a strong view.

@Caslinovich, @JackRonan, @GurliGebis, @YasM0, @ThomasMeyer1970, @Sisku — and anyone I've missed, which is exactly why I'm asking rather than writing.

Sisku:
@Sisku — reported the PIN login bug on issue #4 (phone login always failing with the vehicle-control-PIN error even with the correct PIN; found that email login works around it). Later found that "Refresh full status" succeeds without asking for the PIN when the car is actively charging, but always fails with the PIN error when the car is idle — pointing to the vehicle's deep-sleep state as the real cause, not the PIN itself. Car: Omoda 9. Region: Catalonia, Spain.

JackRonan:
@JackRonan - Worked on extraction of the Omodao Jaecoo APK independently before finding and forking @Caslinovich' work on the omoda9 repo. From there, driving Claude Code through most of the coding, I worked on Omoda/Jaecoo generalisation, PHEV/BEV-agnostic sensors, entity-naming overhaul, the custom Lovelace card, per-vehicle capability detection, charging sensor and re-auth flow, phone/SMS login, and some security hardening. Credit where it's due: Rino had already cracked the auth flow and captcha before I forked, and Luigi pitched in directly too, with PRs for CI hardening, speaking OTP-send errors, and the home/away charging energy feature. Fixes have gone both ways between the two repos afterthat: I ported several parts of Rino's work into the fork, and some of mine went back upstream too. Car: Omoda E5, Region: UK

ThomasMeyer1970:
@ThomasMeyer1970 — field tester. Found that `queryVehicleAuthority` only returns data
[...]

Issue #7, opened 2026-08-19
Private channel for the reverse-engineering material


drake69:
@Caslinovich has the material ready and offered it on 6 August. He'll be ready to hand it over by the end of this month, and then he steps back for a while — so this needs settling this week, not in September.

**It's two different things, and they don't go to the same place.**

The *method* is universal, because the app is universal: how to reach the traffic, the shape of the endpoints, the signing, the MQTT map. That belongs in the repository, in our own words, so anyone with a car can repeat it — it's already started in `docs/field-test.md`.

The *captures* are specific to one car, and they carry decompiled firmware that isn't ours to publish, plus identifiers that are hard to scrub completely. Those stay private, permanently. Not "private until we've cleaned them".

**Proposal:** a private repository in this org, access for whoever is actively working on the protocol. @Caslinovich redacts as he said — VIN, tokens, certificates, account and user ids — and pushes there. If a browser upload is easier than git, that works too; say which you prefer and I'll set it up either way.

Two asks:

- @Caslinovich — which form suits you? That's the only question blocking this.
- Everyone else — say if you want access. It isn't automatic, and there's no expectation that you do.

Nothing from that repository gets copied into this one. What crosses over is what we learn from it, rewritten as rows and as method.

drake69:
The place exists: `chery-connect-ha/protocol-notes`, private, in this org. You can write to it, and so can @JackRonan; @GurliGebis can read it — say the word and that becomes write.

While setting it up I found that the invitations back in August landed as membership rather than ownership, so you two were never actually owners of the org as I said on 4 August. Fixed now, and sorry — it was wrong for the entire time you were deciding whether to hand a repository over.

**How to put things in it**, whichever suits you: `git push`, or **Add file → Upload files** in the browser, which takes about thirty seconds and needs no tooling.

**What it's for:** the knowledge, as text. Endpoints, command schemas and their bodies, the signing notes, the MQTT map, signal dictionaries, redacted excerpts.

**What I'd rather you kept on your own machine:** the APK, the firmware image, whole decompiled trees. Not squeamishness — they're Chery's and Desay's work, a private repo isn't much of a shelter if anyone objects, and git rejects files over 100 MB anyway. If we need something out of them later, we ask you and you look. What we need is the answer, not the file.

[...]

PR #11, opened 2026-08-19
process: how a change lands, and how we say what we tested

drake69:
If you read one thing, read the first page of `CONTRIBUTING.md` — the whole
arrangement is there. The rest is detail you can argue with later.

@Caslinovich asked whether to release the SMS fix directly or open a pull request.
This is my answer, and it comes as a pull request rather than a push to `master` —
which is the first thing it asks for.

Three ideas, and they are the parts worth arguing with:

1. **The gate is proportional.** Data for a car you own: merge. Shared code: one
   read by someone else. Code for a car nobody here owns: merge, labelled
   `unverified-hardware`. A uniform rule prices the harmless change like the
   dangerous one, and then everybody routes around it.

2. **Merging is never blocked; the stable release is.** Merges go out as
   pre-releases, so you have your own work on your own car the same day. A stable
   release needs a field test on each affected model — and may ship unverified
   code, but never silently.

3. **"My approval would be a signature, not a review"** — @Caslinovich, I disagree,
   and this is why. A read of the diff is one kind of review. Running it on your
   Omoda 9 is the other, and it is the one none of the rest of us can do. Half of
   what we ship is written for cars the author does not own; your car is the only
   instrument that can disprove it. `docs/field-test.md` writes down the method you
   were already using — including the part I think is the real finding: an outcome
   is an observation with a date on it, not a property of a model.

One thing I cannot answer for you: when you try something out, do you put it on the
Home Assistant instance you use every day, or somewhere safer? It matters here. The
sandbox in the fork line already exercises the protocol with no instance at all —
the real auth and signing code, live status, commands, raw answers — and it only
works because `core/` has no Home Assistant in it. It should come across with the
port, plus the one thing it lacks: writing its result to a file you can paste into
a pull request instead of reading it off the screen. `docs/field-test.md` says which
claims each instrument can support.

`AGENTS.md` is the practical half: none of us writes this by hand, so the file that
tells our agents how the project works belongs in the repository, with the exact
commands. Nobody should need to learn git to contribute here.

Two notes on scope. The links to `docs/design/` point at #6 and are marked as such,
so this stands on its own if #6 is rejected. And no `CODEOWNERS` yet — the paths it
would name do not exist until the layering lands.

Nothing here is settled and nothing moves before 31 August. If a rule looks like it
would slow you down more than it helps, say so — that is the verdict I want.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,72 @@
+name: Bug report
+description: Something the integration does wrong, or stopped doing
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Before anything else: **remove your VIN, tokens, certificates and account
+        id** from anything you paste. They appear in logs more often than you
+        would expect. `<redacted>` is a perfectly good value.
+
+  - type: dropdown
+    id: model
+    attributes:
+      label: Car
+      description: What this was observed on. It decides who can reproduce it.
+      options:
+        - Omoda 9
+        - Omoda 5 / E5
+        - Omoda 7
+        - Jaecoo 7
+        - Jaecoo 5 / J5
+        - Other Chery-platform car (say which below)
+    validations:
+      required: true
+
+  - type: input
+    id: region
+    attributes:
+      label: Country / region
+      placeholder: IT, UK, DK, …
+    validations:
+      required: true
+
+  - type: input
+    id: versions
+    attributes:
+      label: Integration and Home Assistant versions
+      placeholder: "integration v1.13.0, HA 2026.7.1"
+    validations:
+      required: true
+
+  - type: textarea
+    id: what
+    attributes:
+      label: What you expected, and what happened
+      description: |
+        Times help more than anything else — wall-clock, to the second, for what
+        you sent and for what you saw.
+    validations:
+      required: true
+
+  - type: dropdown
+    id: official_app
+    attributes:
+      label: Does the official app manage it?
+      description: >
+        This separates "the car cannot" from "we ask wrongly", and it is usually
+        the fastest thing in the whole report.
+      options:
+        - "Yes — the official app does it fine"
+        - "No — the official app fails too"
+        - "Not tested"
+    validations:
+      required: true
+
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logs (redacted)
+      description: Debug logs for the integration, with identifiers removed.
+      render: text
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: How this project works
+    url: https://github.com/chery-connect-ha/omoda9-ha/blob/master/CONTRIBUTING.md
+    about: Roles, how a change lands, and why field tests gate the stable release.
+  - name: How to run a field test
+    url: https://github.com/chery-connect-ha/omoda9-ha/blob/master/docs/field-test.md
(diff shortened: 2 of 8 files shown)


GurliGebis:
May I suggest enabling projects in github, and use it to plan and manage a backlog and set up sprints/releases?

That way, it is easy to track how close to a release we are. (I would suggest making a beta release every time a new part of the release is done, so if there are 10 tasks in the sprint for release 5, there would in theory be 10 betas, before the release is finished).

drake69:
@GurliGebis — you proposed this on 20 August and nobody answered you. That is the second time, after the thirteen days on your merged-repo discussion. Here is a real answer, and three things that should have happened earlier.

**Your beta idea already shipped, and it came from you.** You suggested a beta release for each piece of work finished, rather than one at the end. That is now `.github/workflows/beta.yml` (#12, merged this morning): putting the `beta` label on a pull request publishes an installable pre-release, so somebody with the right car can try a change *before* it lands. The first one went out an hour ago — `v1.14.0-beta.1`, for a phone-login fix that the person who wrote it cannot test, because the bug only affects accounts with no e-mail address.

I arrived at it from the opposite direction — the review rule was unenforceable because a pull request produced nothing anyone could install — and you got to the same missing piece from sprints and releases. You were right about the half that matters.

**On Projects and sprints, a real position rather than a thank-you.**

**Yes to a board.** There are eleven open pull requests right now and no single place that shows what is in flight, what is waiting on a person, and what is waiting on a car. #10 holds the *plan*, deliberately edited in place rather than re-explained, but it is a sequence and not a state. That gap is real and you are the one who named it.

**No to sprints with dates**, and the reason is about us rather than about the method: nobody here can commit to a date. @Caslinovich is away from early September, @JackRonan lost his machine for two weeks, I went quiet for eleven days in August. A sprint that nobody can hold turns into a fiction we all stop reading, and this project already gates on evidence — a change becomes stable when somebody's car has done the thing — which is a condition, not a calendar.

**So: it is yours if you want it.** I have enabled Projects on this repository (it was off, which is why your suggestion could not have gone anywhere even if someone had replied). You now have write access. Set it up the way you think it should work — I would rather have your shape for it than mine, and if you make a board and it turns out to be overhead, we drop it and nothing is lost.

**Three things that were wrong, stated plainly rather than apologised for.**

1. **The idea of combining the two repositories was yours**, on 19 July — nine days before I proposed it. Your issue is literally titled "Why not combine efforts?".
2. **You built the merged repository** and asked for a review on 6 August. You got one on the 19th.
3. **Until today you had read-only access to this repository** — the one you were being asked to merge into, and the one where you offered to do the HACS default-store pull request. You now have write. That was an oversight of the same kind as the invitations that landed as membership instead of ownership back in August, and it explains more about the last three weeks than any theory about anybody's interest.

**Two asks, both genuinely for you.**

**One technical, on #19.** I labelled two pull requests `beta` at the same moment. One produced a pre-release; the other produced no workflow run at all and no message saying why. `beta.yml` is absent from *both* branches, so that is not the difference; the only thing I can see is that the silent one has not been pushed since before the workflow existed on `master`. I do not know the mechanism and I would rather be told than keep guessing — you work with Actions more than any of us.

**One about the file.** `CONTRIBUTORS.md` is in #23, written from what people wrote about themselves on #8. Yours is missing because you have not written it and I am not going to put words in your mouth. If you would rather I drafted something for you to correct, say so — otherwise the file names you under "still missing" with one clause about the merged repository and the HACS offer, which is less than you did.

*Attribution, per the convention on #7: the checks behind this — the label behaviour, the repository settings, the dates — were run by the agent I drive. The position on sprints, and the decision to hand the board to you rather than build one, are mine.*

GurliGebis:
Thank you - I'll take a look at setting up a project, backlog etc. soon (next few days will be busy) 😊

PR #12, opened 2026-08-22
ci: a beta build for every pull request, installable from HACS

drake69:
Labelling a pull request `beta` publishes a pre-release anyone can install from HACS. It is the missing half of the rule this project has already agreed on.

## Why

The gate here is the **stable release**, not the merge, and the review that counts is someone running a change on their own car — half of what we ship is written for cars the author has never sat in, and on the author's own car it is a no-op that cannot be falsified.

That rule was not exigible, because **a pull request produced nothing anybody could install**. `hacs.json` sets `zip_release` with `omoda9.zip`, so HACS installs only from a release carrying that asset, and the asset is built and uploaded by hand, once per release. The only way to field-test a change before it landed was to download a branch and copy files into `custom_components/` by hand — precisely the barrier that filters out the people whose review is worth the most.

@GurliGebis proposed a beta per unit of work on #11, coming at it from sprints. Same missing piece.

## How it works

**Put the label `beta` on a pull request.** A label is two taps in a phone browser, which is deliberate: whoever needs to ask for a field test does not always have a computer in front of them. `workflow_dispatch` is there as an escape hatch.

Only people with write access can label, so only they can cut a beta. Testers never trigger anything — they get a link and install.

The release is tagged `v<manifest version>-beta.<n>`, where `<n>` is the first free counter across existing tags, so two pull requests aiming at the same version cannot collide. Which pull request a beta came from is in the release **title**, not the tag, so the tag stays clean semver and HACS orders it correctly.

When a pull request that touches `custom_components/` is opened, the workflow leaves one comment saying the label exists. A line in `CONTRIBUTING.md` is not read by whoever opens a pull request in three weeks; this announces itself at the moment it is needed. Documentation-only pull requests and bots get nothing.

## Two things done on purpose

**Never from a fork.** That zip is installed into Home Assistant instances holding somebody's car credentials, and published from the canonical repository it looks official. An outside contributor must not be able to get an artifact out with that stamp. From a fork, a member cuts the beta by hand after reading what is in it.

**The beta version is stamped into the manifest inside the zip**, so a tester can see which build is running rather than taking our word for it.

## What was verified, and what was not

**Verified.** The archive is built from tracked files only and compared entry by entry against the `omoda9.zip` of v1.13.0: same 53 files, same layout — the contents of `custom_components/omoda9/` flattened to the archive root, which is what HACS expects with `content_in_root: false`. The YAML parses and the shell steps are shellcheck-clean.

**Not verified, and this is the part that needs somebody with a car.** I have not seen HACS list this pre-release and install it. That requires a Home Assistant instance, and until someone does it, this is "the backend accepted" and not "the car did it" — the exact distinction `AGENTS.md` asks us not to blur. **The first beta this cuts is its own field test**, and whoever installs it first should say whether it appeared under *Show beta versions* and whether the install completed.

## Related

Documentation for this lands in #11, which also carries a fix: its *Cutting a pre-release* snippet says `gh release create v<X.Y.Z> --prerelease --generate-notes`, which produces a **broken** beta — with `zip_release` set and no asset attached, HACS lists it and the install fails.

Follow-up worth doing separately: stable releases still have the zip built by hand, with the same trap.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -1,0 +1,216 @@
+name: Beta
+
+# Una pre-release INSTALLABILE tagliata da una pull request, per farla provare su
+# un'auto vera prima che il cambiamento entri.
+#
+# Perché esiste. Il gate di questo progetto non è il merge ma la STABILE, e la
+# review di chi non programma è la prova sul campo sulla propria auto. Perché
+# quella regola sia esigibile, chi ha l'auto deve poter installare la PR — e fino
+# a oggi non poteva: `hacs.json` dichiara `zip_release`, quindi HACS installa solo
+# da una release che porta allegato `omoda9.zip`, e quell'asset veniva costruito e
+# caricato a mano una volta per release. L'unica strada era scaricare un branch e
+# copiare i file in `custom_components/` a mano: esattamente la barriera che tiene
+# fuori le persone la cui review vale di più.
+#
+# Come si avvia: si mette l'etichetta `beta` sulla PR. Due tocchi dal browser di un
+# telefono — è deliberato, perché chi deve poterla chiedere non sempre ha un
+# computer sotto mano. `workflow_dispatch` resta come uscita di sicurezza.
+#
+# Chi può avviarla: solo chi ha accesso in scrittura, perché è GitHub a decidere
+# chi può etichettare. I tester non avviano niente: ricevono un link e installano.
+
+on:
+  pull_request:
+    types: [opened, labeled]
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: "Numero della PR da cui tagliare la beta"
+        required: true
+        type: string
+
+permissions:
+  contents: write        # creare il tag e la pre-release
+  pull-requests: write   # scrivere il commento sulla PR
+
+jobs:
+  # Il meccanismo si annuncia da solo nel momento in cui serve. Una riga in
+  # CONTRIBUTING.md non la rilegge nessuno: chi apre una PR fra due settimane non
+  # saprebbe che la beta si può chiedere, e saremmo di nuovo senza tester.
+  suggerimento:
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.action == 'opened' &&
+      github.event.pull_request.user.type != 'Bot'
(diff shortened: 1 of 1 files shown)


drake69:
Evidence from this pull request itself, since it is the first thing the workflow ran on.

Both guards behaved: `suggerimento` ran and deliberately posted nothing — this pull request touches only `.github/`, so the documentation-only check held — and `beta` skipped, there being no label. The three required checks are green.

What that establishes is that the workflow parses, triggers on the right events and declines on the wrong ones. It says nothing yet about the part that matters: **nobody has seen HACS list one of these pre-releases and install it.** That needs a Home Assistant instance. The first beta cut from a real pull request is the field test of this one, and whoever installs it should say whether it appeared under *Show beta versions* and whether the install completed.

drake69:
Merged, along with #11 which carries the documentation.

**@JackRonan — this is now available to you immediately.** On #7 you asked @divisl to download a branch zip and extract it into `custom_components/` by hand. For anything with an open pull request that is no longer necessary: put the label `beta` on it and a pre-release comes out, installable from HACS with *Show beta versions* on. Two taps, and the tester never has to touch a filesystem. That applies to #16, #17 and #18 right now.

One thing it cannot reach: `feat/brand-region-presets` has no pull request, so it has no CI and no beta. If you open it as one — even as a draft — divisl gets an installable build of the Chery endpoints instead of a manual extraction, and you get the checks.

**On how this merged, since I am asking the rest of you to follow the rule I just wrote.** It went in **without a non-author read**, and not by way of the one-week valve in `CONTRIBUTING.md` — that would be five days early. I judged it proportional: no runtime code changes, nothing reaches a user until somebody chooses to label a pull request, and the alternative was leaving the mechanism unavailable on the day it became useful.

What that judgement does **not** cover is the part worth reading: the **fork guard**. That zip is installed into instances holding people's car credentials, and published from this repository it looks official, so a beta must never be cut from a fork. The guard is doubled — the job `if` for the label event, plus an explicit check inside the step for the `workflow_dispatch` path, where there is no pull-request event to read the head repository from.

I would rather have that read **before the first beta is cut than before the merge**. If you or @Caslinovich look at nothing else in it, look at that.

PR #14, opened 2026-08-24
Fix SMS/mobile login: identity order + clearer no-vehicle error

### LINKED ISSUE #5: Login via SMS: il BFF risponde app.service.failed / "contatta il servizio clienti" (account registrato solo col numero)
### Ambiente

- Integrazione **v1.13.0**, installata a mano in `custom_components/`
- Home Assistant **2026.8.1** su HA OS (VM su Proxmox)
- Veicolo: **Jaecoo 7 PHEV**, Italia
- Account Omoda/Jaecoo **registrato solo col numero di telefono**: non ha un indirizzo email associato, quindi l'accesso via email non è percorribile
- L'**app ufficiale funziona regolarmente** con lo stesso account

### Cosa succede

Il config flow via SMS arriva fino in fondo e poi si ferma con:

```
Motivo: token coniato ma login ancora KO
Codice OTP non valido o scaduto
```

Il secondo messaggio è fuorviante: il codice OTP è corretto. Da `session.confirm_otp()` si vede che quel ramo è raggiunto **solo se `prova_token.py` ha restituito `RESULT: OK`** — quindi il token è stato coniato — ed è `check()` a fallire subito dopo.

Log con `custom_components.omoda9: debug`:

```
DEBUG  Omoda9 login: invio OTP (SMS) rc=0
DEBUG  Omoda9 login: conio token (SMS) rc=0
       [login mobile mobile=APP-LOGIN@***XXXX_39 code_len=6 code_fmt=plain loc=BODY] HTTP 200
       resp: {"code": 0, "msg": "operation.successful", ... "access_token": "<128ch redatto>",
              "refresh_token": "<128ch redatto>", "client_id": "legendApp", ...}
       LOGIN OK - token salvato in /config/omoda9_pending_token.json
       RESULT: OK
WARNING Omoda9: conferma OTP fallita: token coniato ma login ancora KO
```

Quindi `/auth/oauth2/token` risponde `code: 0, operation.successful` e conia access_token + refresh_token regolari.

### La risposta del BFF

`_bff_login()` restituisce `(None, None)` senza registrare nulla, quindi ho aggiunto una riga di log temporanea subito dopo la `requests.post` in `core/wake.py` per vedere la risposta vera:

```
OMODA9-DIAG bff_login -> HTTP 200
  url=https://legend-oj.omodaauto.nl/api/tsp/v1/app/auth/login
  dept=39  channelId=1  countryId=1  tenant=300006
  body={"code":1,
        "msg":"Per favore contatta il servizio clienti per la gestione",
        "key":"app.service.failed",
        "data":null,
        "remarks":null,
        "txId":null,
        "ok":false}
```

Riprodotto **identico due volte** (il secondo è il ritentativo dopo `_refresh_token`).

### Perché mi sembra rilevante

- **HTTP 200 con `code: 1`**: non è un rifiuto di autenticazione né un token scaduto. Il BFF accetta il `Bearer`, elabora la richiesta e la respinge a livello applicativo.
- Il messaggio arriva **localizzato in italiano**, coerente con `DEPT-ID: 39` — la regione viene interpretata correttamente.
- I parametri di regione sono tutti quelli predefiniti e corretti per l'Italia: `dept=39`, `channelId=1`, `countryId=1`, `tenant=300006`.
- `key: app.service.failed` è un errore di *servizio*, non di credenziali.

L'ipotesi che mi pare più coerente — ma è un'ipotesi, non una misura — è che il BFF non riesca a risolvere l'identità composita `APP-LOGIN@<numero>_<area>` in un profilo TSP abilitato, mentre con l'identità email la risoluzione va a buon fine. Combacia con quanto riportato in #4, dove l'accesso email funziona e quello SMS no.

L'app ufficiale, con lo stesso account e lo stesso numero, entra senza problemi: quindi l'account **è** abilitato lato Chery, e la differenza sta in come viene costruita l'identità o in qualche parametro che l'app manda e la ricostruzione no.

### Cosa posso fare per aiutare

Ho l'ambiente pronto e posso rifare la prova con parametri diversi, se indichi cosa provare — per esempio un `channelId` alternativo, un `countryId` diverso, o un formato diverso dell'identità nel body. Posso anche fornire il log completo con `debug` attivo.

La riga di diagnostica che ho usato è questa, subito dopo la `requests.post` in `_bff_login`:

```python
_LOGGER.warning("OMODA9-DIAG bff_login -> HTTP %s | url=%s | dept=%s ch=%s country=%s tenant=%s | body=%s",
                r.status_code, ctx.bff + "/tsp/v1/app/auth/login",
                H.get("DEPT-ID"), H.get("channelId"), H.get("countryId"), H.get("TENANT-CODE"),
                (r.text or "")[:400])
```

Forse varrebbe la pena tenerla stabilmente a `debug`: oggi un fallimento del BFF è indistinguibile da un OTP sbagliato, e l'utente viene indirizzato a richiedere codici che non servono.

Grazie per il lavoro sul progetto.

JackRonan:
Fixes #5

Two related fixes to the SMS/mobile login experience.

## 1. Correct the SMS-login composite identity order

Mobile (SMS) login built the OAuth identity as `APP-LOGIN@<number>_<areaCode>`, but the backend
expects `APP-LOGIN@<areaCode>_<number>`. Reversed, the token was minted against a **phantom**
account (a user record with `areaCode`/`phone` swapped, no vehicles, no TSP user) instead of the
caller's real account — so `getTuserId` returned `"0"` and `login` returned `app.service.failed`,
and **every** SMS login failed at vehicle discovery even though OTP send and token mint succeeded.

Verified live by comparing email and SMS logins of the same account: the corrected order resolves
to the real account (same `tUserId` and vehicle as the email login); the old order resolved to the
empty phantom. Also updates `mask.identita_mobile` to redact the **second** field (the number)
rather than the first — with the corrected order the first field is only the dialling prefix, so
the old masking would have leaked the number into logs.

## 2. Show "no vehicle" instead of "invalid OTP" when the code was valid

When the OTP is correct and the token mints, but the account has **no vehicle**, the post-mint
login check fails and `async_step_otp` used to label it **"invalid OTP code"** — sending users to
re-check a code that was fine. The usual trigger is signing into the **wrong account** (e.g. the
phone number when the car is on the email login).

`async_step_otp` now routes on whether a token was **actually minted** (`_pending_token_minted`):
if one was written it proceeds to discovery, which correctly reports **`no_vehicle`**; only a
genuinely rejected code shows `otp_invalid`. The `no_vehicle` message now hints to sign in with the
other login (e.g. email). Translations updated (en / it / strings).

No behaviour change for email login; existing entries unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ dell'integrazione: aggiorna da **HACS → Omoda 9 / Jaecoo → Aggiorna**.
 
 ## [Non rilasciato]
 
+## v1.14.0 — 2026-08-24
+
+### 🇮🇹 Italiano
+
+- **Il login via SMS ora funziona, anche per gli account registrati solo col numero.** Chi accedeva col numero di telefono invece che con l'email otteneva sempre un errore, anche col codice giusto: l'integrazione componeva l'identità dell'account **nell'ordine sbagliato** (prefisso e numero invertiti) e coniava un token per un account fantasma vuoto, senza veicoli. Ora l'ordine è quello corretto — confermato sia dal vivo sia leggendo il programma dell'app ufficiale — e il login via SMS raggiunge il tuo account reale, con la tua auto.
+- **Un messaggio chiaro quando l'account non ha un'auto.** Se il codice era corretto ma su quell'account non risulta nessun veicolo, prima compariva «codice non valido», mandandoti a ricontrollare un OTP che invece andava benissimo. Adesso te lo diciamo com'è davvero: il codice è giusto, ma non c'è un'auto collegata a quell'account.
+- **Numero di telefono più protetto nei file di diagnostica.** La regola che nasconde il numero nei log è stata resa più robusta, così non può sfuggire per un caso di forma inattesa: nel dubbio nasconde di più, mai di meno.
+
+### 🇬🇧 English
+
+- **SMS sign-in now works, including for accounts registered with a phone number only.** Signing in with a phone number instead of an email always failed, even with the right code: the integration built the account identity **in the wrong order** (area code and number swapped) and minted a token for an empty phantom account with no vehicles. The order is now correct — confirmed both live and by reading the official app — so SMS sign-in reaches your real account and your car.
+- **A clear message when the account has no car.** If the code was correct but that account has no vehicle on it, it used to say "invalid code", sending you to re-check an OTP that was actually fine. Now it tells you what is really going on: the code is right, but there is no car linked to that account.
+- **Safer phone-number redaction in the diagnostics file.** The rule that hides the number in logs was hardened so it cannot slip through on an unexpected shape: when in doubt it masks more, never less.
+
 ## v1.13.0 — 2026-08-10
 
 ### 🇮🇹 Italiano
--- a/README.it.md
+++ b/README.it.md
@@ -195,7 +195,7 @@ o telefono) + OTP. Catena orchestrata dal config flow (codice in
 |---|---|---|
 | invio OTP (email) | `login_omoda.py invia <email>` | risolve il captcha del gateway (§2) e fa partire il codice via **email** |
 | invio OTP (SMS) | `login_omoda.py invia-sms <numero-senza-prefisso> <prefisso>` | idem via `sendSmsCode` — l'unico endpoint dietro un WAF Aliyun che filtra sull'**impronta TLS**, gestito da `tls_client.py` |
-| conio token | `prova_token.py <email> <code>` | chiama `/auth/oauth2/token` replicando l'app (cifratura SM4) e salva il token. Per gli account col numero l'identità è la composita `APP-LOGIN@<numero>_<area>` |
+| conio token | `prova_token.py <email> <code>` | chiama `/auth/oauth2/token` replicando l'app (cifratura SM4) e salva il token. Per gli account col numero l'identità è la composita `APP-LOGIN@<area>_<numero>` (prima il prefisso, poi il numero — confermato in `UserService::phoneVerifyLogin`) |
 | orchestrazione | `session.py` | espone `request_otp()` / `confirm_otp(code)` / `check()` / `refresh()` |
 
 Il **PIN non c'entra con l'accesso**: è l'OTP a coniare il token, il PIN firma
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ phone) + OTP. Chain orchestrated by the config flow (code in
 |---|---|---|
 | send OTP (email) | `login_omoda.py invia <email>` | solves the gateway captcha (§2) and triggers the code by **email** |
 | send OTP (SMS) | `login_omoda.py invia-sms <number-without-country-code> <country-code>` | same, via `sendSmsCode` — the one endpoint behind an Aliyun WAF that filters on the **TLS fingerprint**, handled by `tls_client.py` |
-| mint token | `prova_token.py <email> <code>` | calls `/auth/oauth2/token` replicating the app (SM4 encryption) and saves the token. For phone accounts the identity is the composite `APP-LOGIN@<number>_<area>` |
+| mint token | `prova_token.py <email> <code>` | calls `/auth/oauth2/token` replicating the app (SM4 encryption) and saves the token. For phone accounts the identity is the composite `APP-LOGIN@<area>_<number>` (area code first, then number — confirmed in `UserService::phoneVerifyLogin`) |
 | orchestration | `session.py` | exposes `request_otp()` / `confirm_otp(code)` / `check()` / `refresh()` |
 
 The **PIN plays no part in signing in**: the OTP mints the token, the PIN only
--- a/custom_components/omoda9/config_flow.py
+++ b/custom_components/omoda9/config_flow.py
@@ -149,11 +149,11 @@ _MAX_CIFRE_PREFISSO = 3          # E.164: i prefissi paese hanno 1, 2 o 3 cifre.
 
 def _normalizza_telefono(numero: str | None, prefisso: str | None) -> tuple[str, str]:
     """Ripulisce numero e prefisso UNA VOLTA SOLA, qui: da qui in poi viaggiano in entry.data,
-    nell'ambiente dei sottoprocessi e nell'identità `APP-LOGIN@<num>_<area>`, e a valle nessuno
+    nell'ambiente dei sottoprocessi e nell'identità `APP-LOGIN@<area>_<num>`, e a valle nessuno
     li tocca più (`invia_sms` fa solo `lstrip("+")` + spazi, `build_params_mobile` altrettanto).
 
     Cose che la gente scrive davvero e che senza questo passaggio arrivano storte al server:
-      * numero copiato dalla rubrica col prefisso già dentro → identità `APP-LOGIN@39<num>_39`;
+      * numero copiato dalla rubrica col prefisso già dentro → identità `APP-LOGIN@39_39<num>`;
       * separatori dentro il numero: spazi, punti, trattini;
       * prefisso scritto `+39` o `0039` → `areaCode="+39"`, che il server non riconosce;
       * zero di accesso nazionale davanti al numero (Regno Unito, Germania, Francia…).
--- a/custom_components/omoda9/config_flow.py
+++ b/custom_components/omoda9/config_flow.py
@@ -64,6 +64,22 @@ def _pending_token_path(hass: HomeAssistant) -> str:
     return hass.config.path(f"{DOMAIN}_pending_token.json")
 
 
+def _pending_token_minted(hass: HomeAssistant) -> bool:
+    """True se un token è stato davvero SCRITTO (OTP valido), anche quando `confirm_otp` ha poi
+    riportato un fallimento perché il login post-conio non è riuscito — di norma perché l'account
+    NON ha veicoli. Serve a distinguere «codice sbagliato» (nessun token) da «codice giusto ma
+    niente auto sull'account»: il primo è `otp_invalid`, il secondo `no_vehicle`, e finché non li
+    si separava un login sull'account sbagliato usciva come «OTP non valido», mandando l'utente a
+    ricontrollare all'infinito un codice che era corretto."""
+    try:
+        with open(_pending_token_path(hass), encoding="utf-8") as fh:
+            tok = json.load(fh)
+    except Exception:  # noqa: BLE001
+        return False
+    d = tok.get("data", tok) if isinstance(tok, dict) else {}
+    return bool(d.get("access_token") if isinstance(d, dict) else None)
+
+
 def _reason_line(detail: str | None) -> str:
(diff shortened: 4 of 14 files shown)


JackRonan:
Closing - had some unforeseen issues in my testing

Caslinovich:
Hi Jack,

First things first: this is a bug I shipped in v1.8.0 and could never have found. My account
has an e-mail, so everything past the mint always worked here. You found it and fixed it —
thank you. I've read the reopened version, and the second half is the right call too.

**One thing I can add that your setup couldn't: the field order is also readable in the
decompiled app**, so it no longer rests only on a live comparison. In
`UserService::phoneVerifyLogin` the identity is built as `APP-LOGIN@{arg2}_{arg3}`, and which is
which resolves one call down: `isPhoneNumInValid(arg2, arg3)` shows the "enter your mobile
number" toast when **`arg3`** is empty, then calls `isValidPhoneNum2(receiver = arg3,
argument = arg2)` — and inside that, the argument goes to `AreaString.phoneMaxInputLength(...)`,
while the receiver goes to `splitPhoneNumber`, which strips that prefix off its front. So `arg2`
is the area code and `arg3` the number: `APP-LOGIN@<areaCode>_<number>`, exactly as you have it.

For the record, what let the wrong order survive on my side wasn't the decompiler. The wrong
order *mints a valid token*, and I took that as end-to-end confirmation. A
`200 operation.successful` only says the request was well formed — not whose account it reached.

**Cross-checked on the Omoda/Jaecoo gateway.** Your debugging was on a Chery EU account, so I
ran the same thing on a number-only account here: with the corrected order,
`GET /marketing/v1/app/user` returns my own `phone` and `areaCode` in the right fields —
alongside the `user/signOffInfo` / `user/detailById` reads you already used, this one names the
identity explicitly. Same three answers you documented in #5 otherwise (`tUserId "0"`,
`app.service.failed`, empty `queryList`), on an account with no car.

**One thing I'd change in `mask.py`.** Nothing leaks in practice — neither string below can come
out of `build_params_mobile`. But swapping *which* field is redacted keeps the masking
positional, and position is the assumption that produced this bug:

```
APP-LOGIN@<number>_39   ->  APP-LOGIN@<number>_***     # old-order string, whole number
mario@rossi.it_39       ->  rossi.it_***               # master masks this one
```

The second is a real regression and the suite misses it: the existing test only forbids the
string `mario`, not the domain. I'd print the first field only when it looks like a dialling
prefix — 1 to 3 digits, the `config_flow._MAX_CIFRE_PREFISSO` bound we already state — and mask
both otherwise, so the only way it can err is masking too much.

**CI.** The three failures are just tests still encoding the old order
(`test_identita_mobile_e_composita`, `test_identita_tiene_la_forma_e_butta_il_numero`,
`test_identita_di_forma_inattesa_non_tira_a_indovinare`), plus the two READMEs. I have that
realignment green locally, with the `mask.py` guard kept separate so it can't hold this up.
You're on a borrowed laptop — unless you'd rather do it yourself, say the word and I'll push it
to your branch.

Once CI is green and it's out of draft, it goes in as far as I'm concerned.

**@MassimoDe** — short version: we've found it, and it's a bug in code I wrote, not anything
wrong with your account. Two honest caveats. Nobody has yet tested this exact case, an account
with a car and no e-mail, which is yours — that's the one piece of evidence none of us can
produce. And when you do try it, don't judge it by `tUserId`: as Jack found, `"0"` also means
"no TSP user mapped", so what tells you it worked is the car appearing in `queryList`. We can
now put a `beta` label on this PR and have it build an installable pre-release, so you won't
have to copy files by hand — I'll ping you here when there's one.

Nothing to do on an existing install in the meantime: this only affects phone/SMS sign-in, which has been
broken since it shipped in v1.8.0; e-mail sign-in and already-configured entries are untouched.

— Rino

JackRonan:
@Caslinovich — thanks, that's a great catch and I've taken all of it on board. Pushed as `705ff04`:

- **`mask.py` guard** exactly as you suggested: `identita_mobile` now reprints the first field only when it's genuinely a dialling prefix (`_e_prefisso`: 1–3 digits, the same `config_flow._MAX_CIFRE_PREFISSO` bound). A domain, a name, or a number that lands there by mistake is masked — so the only way it can err is masking too much.
  - `mario@rossi.it_39` → `***_***` (domain gone — the regression you spotted)
  - `APP-LOGIN@3001234567_39` (inverted order) → `APP-LOGIN@***4567_***` (no full-number leak)
  - `APP-LOGIN@39_3001234567` → `APP-LOGIN@39_***4567` (normal, unchanged)
- **Tests realigned** to the correct order (`test_identita_mobile_e_composita`, `test_identita_tiene_la_forma_e_butta_il_numero`, `test_identita_di_forma_inattesa_non_tira_a_indovinare`), plus a new `test_identita_maschera_il_primo_campo_se_non_e_un_prefisso` that forbids the leaked **domain**, not just `mario`.
- **Docs**: both READMEs and the stray `<num>_<area>` references fixed, and I folded your decompiler evidence (`UserService::phoneVerifyLogin` → `APP-LOGIN@{areaCode}_{number}`, resolved via `isValidPhoneNum2`) into the `build_params_mobile` docstring — it no longer rests only on the live comparison. 

Would you be able to give it a review? If it looks right to you and CI comes back green, happy to proceed however you'd prefer on the merge — take it out of draft and, if you still want to, put the `beta` label on so @MassimoDe gets an installable pre-release to try the car+no-email case before it lands.

drake69:
@JackRonan @Caslinovich — you both reached for the `beta` label, and two things were in the way. Both are handled, one needs a line from you.

*Per the convention on #7: the checks below were run by the agent I drive; the calls are mine.*

**1. The label did not exist.** `CONTRIBUTING.md` and `AGENTS.md` tell you to use `beta`, `unverified-hardware` and `merged-unread`, and I merged those documents this morning without ever creating the labels. Applying one would have given you an error on first use. All three exist now. That one is on me.

**2. The manifest needs a bump before the beta will build**, and this is the part that needs you, @JackRonan.

This branch declares `version: 1.13.0`, which is also the current stable. The tag would be `v1.13.0-beta.1` — and in semver a pre-release sorts **before** the version it is named after, so HACS would list it as *older* than the 1.13.0 @MassimoDe already has installed. It would never appear as an update, and selecting it would formally be a downgrade. A beta that looks fine and cannot work.

So: **set `version` to `1.14.0` in `custom_components/omoda9/manifest.json`, commit to this branch, then apply the label.** One line, and it also decides the number this ships under, which is a decision rather than a formality.

I have opened **#22** so the workflow refuses that tag instead of publishing it silently. It names the number to bump to rather than picking one — guessing would mean publishing a beta under a number different from the stable that follows it.

**@Caslinovich — on the offer to push the CI realignment to Jack's branch:** he has already pushed `705ff04` and CI is green on all three checks. So that one is closed.

**@MassimoDe** — Rino has told you the important part already: it is a bug in our code, not something wrong with your account, and the thing that tells you it worked is your car appearing in `queryList`, not `tUserId`. Two additions from me.

**What happens next, so you are not waiting on a silence.** Once the version line above lands, this pull request gets the `beta` label and a pre-release appears on the releases page. In HACS: the integration's ⋮ menu → *Redownload* → turn on **Show beta versions** → pick `1.14.0-beta.1` → restart. To go back, the same path and pick the last stable.

**And if it does not work, that is a result and not a failure — please say so.** You are the only person who can test this exact case: an account with a car and no e-mail address. None of the four of us can produce that evidence. If the car does not appear, the useful thing to post is what the config flow said and at what step it stopped, rather than "still broken" — that is what separates a wrong fix from a second, different bug sitting behind the first.

**One caveat you are owed:** this is a pre-release, and the account it touches is the one that controls your car. If you have a spare Home Assistant instance, use that.

JackRonan:
Done — `manifest.json` is now `version: 1.14.0` (`4a8bcc3`), so the pre-release tags as `v1.14.0-beta.1` and sorts *after* the current stable rather than looking like a downgrade. Thanks for creating the labels and for the #22 guard.

Ready for the `beta` label whenever you are — I'll leave applying it (and cutting the pre-release) to you so the publish is a deliberate step. Happy for this to ship as 1.14.0.

Caslinovich:
@JackRonan — yes, that closes what I raised.

*Per the convention on #7: the run below, the call-site search and the reading of
`prova_token.py` were done by the agent I drive. I have not read the diff myself. The
judgement is mine.*

**The guard behaves.** It was run against your three cases and five more, taken from your
branch. Using the suite's own placeholder number, the outputs are:

```
APP-LOGIN@<area>_<number>       ->  APP-LOGIN@39_***4567     (normal)
APP-LOGIN@<number>_<area>       ->  APP-LOGIN@***4567_***    (inverted: no leak)
<the malformed case in test_mask.py>  ->  ***_***            (the regression I spotted)
APP-LOGIN@<area>_<2 digits>     ->  APP-LOGIN@39_***
APP-LOGIN@_<number>             ->  APP-LOGIN@***_***4567
APP-LOGIN@<area>                ->  APP-LOGIN@***
```

All eight reproduce byte for byte. The one thing that could not be told from the description:
**`identita_mobile` has exactly one call site in the integration** (`core/prova_token.py`), so
the guard is not being bypassed somewhere else. That was the part worth checking, and it holds.

One caveat on this comment rather than on your code: whether a run by an agent counts as the
non-author read `CONTRIBUTING.md` asks for on `core/` is not something I should settle by
writing as if it did. Treat this as evidence, and let somebody decide whether it is also the read.

**On the red CI.** The only failing test is `test_versione_manifest_e_changelog_allineati`:
the manifest declares 1.14.0 and the CHANGELOG has no `## v1.14.0` section. It is a hygiene
test and says nothing about the behaviour of the code — but `AGENTS.md` wants the suite green
in the pull request that changes the behaviour, so it has to go before this lands. (Nobody has
installed the pre-release yet, so I am not going to tell anyone it is safe.)

**I am happy to write that CHANGELOG section if you would rather not.** It has to be bilingual
IT/EN — HACS shows one text to everybody — and in plain language for people who do not read
code. That is the one habit from my side of this project I would like to keep. Say the word and
I will push it to your branch, or paste it here for you.

Separately, and **not for this branch**: in that same function the e-mail path still puts the
identity in the query string, and the `post` is outside the `try` — the same shape @drake69
described on #16. It is my code, from before your fork, and one behaviour per pull request. My
mistake, my issue to open.

— Rino

drake69:
Conflict resolved and pushed to this branch — `master` merged in, your commits untouched.

**What the conflict was:** `CHANGELOG.md` only. Both this branch and `master` wrote into the
same `## v1.14.0` section, which is the fallout of the version rule we introduced on 25 August
(#46), not of anything wrong here.

**How it was resolved:** both sets of entries kept, `master`'s first, in both languages —
5 Italian bullets and 5 English, no text rewritten. `manifest.json` now matches `master`
exactly, so this PR no longer bumps the version, as the rule asks.

**Verification — this part was done by the agent, not read line by line by me:**
full suite on the merge result, **649 passed**, on Python 3.13 as in CI.

Nothing here is a field test: the SMS-login path still needs someone whose account is
registered with a phone number only.

PR #16, opened 2026-08-24
Add username + password login (no OTP)

### LINKED ISSUE #13: Add a way to log with login and password
Official app already suport it, it is way easier no need for codes or sms. Any chance we could have it?

JackRonan:
Closes #13

Adds a third sign-in method — **"Sign in with password"** — next to the email-OTP and SMS-OTP
flows. It uses the backend's OAuth **ROPC** path (`grant_type=password` with a raw username +
password), so there's **no OTP code and no captcha**, matching what the official app offers.

### Behaviour
- New menu option in setup: enter **email + account password + vehicle PIN** (+ region fields), and
  the VIN/account are discovered automatically — no code step.
- **Mint once, ride the refresh token.** The password is used a single time to mint the token; from
  there the session lives on the `refresh_token`, exactly like the OTP flow.
- **Reauth** for password accounts re-prompts for the password (a `login_method` marker on the entry
  drives this) instead of sending an OTP.

### Security
- The password is a **one-shot credential**: it is **never written to `entry.data`** and **never
  logged** (only its length appears in debug output).
- It reaches the mint subprocess **only via the ephemeral environment** (never argv, so it can't leak
  via `ps`/`/proc`), the same channel the OTP code already uses.
- If both access and refresh tokens expire, reauth asks for the password again rather than storing it.

### Changes
- `core/prova_token.py`: `build_params_password` / `call_password` + password branch in `main()`.
- `core/session.py`: `login_with_password`.
- `coordinator.py`: `_login_with_password` (reauth).
- `config_flow.py`: `login_password` step + `reauth_password` step.
- `const.py`: `CONF_PASSWORD` (transient), `CONF_LOGIN_METHOD`.

The core ROPC mint is verified live (returns a valid token on a real account). The config-flow UI
wants a quick end-to-end check in HA — hence draft.

### DIFF
--- a/.code_reader/deliveries/2026-08-19_architecture-seed-v113.json
+++ b/.code_reader/deliveries/2026-08-19_architecture-seed-v113.json
@@ -1,0 +1,47 @@
+{
+  "delivery_id": "2026-08-19_architecture-seed-v113",
+  "date": "2026-08-19",
+  "repo": "chery-connect-ha/omoda9-ha",
+  "intent": "Pubblicare il seme architetturale sul repo canonico: vocabolario e invarianti del dominio, rotta in cinque fette rilasciabili, e un gate CI a cricchetto che nasce con le sole regole gia' vere. Nessun codice runtime toccato.",
+  "baseline": "29fef39 (v1.13.0)",
+  "head": "b7d990f",
+  "delivered_as": "PR chery-connect-ha/omoda9-ha#6 (6/6 check verdi)",
+  "units": [
+    {
+      "block": "intent",
+      "path": "docs/design/domain-model.md",
+      "lines": 227,
+      "note": "Vocabolario (capability, placeholder frame, backend acceptance, charge session, zone) e invarianti. La tabella invariante/comportamento-attuale/sintomo e' la parte che porta il peso: i quattro bug aperti sono quattro violazioni. Include la nota su cosa v1.10.0-v1.13.0 ha gia' confermato.",
+      "reading_estimate_minutes": 14
+    },
+    {
+      "block": "intent",
+      "path": "docs/design/architecture.md",
+      "lines": 159,
+      "note": "Mappatura modulo-per-modulo verso transport/, platform/, domain/, features/ e le cinque fette. Prevalentemente ri-seduta, non riscrittura.",
+      "reading_estimate_minutes": 9
+    },
+    {
+      "block": "test",
+      "path": "tests/test_architecture.py",
+      "lines": 182,
+      "note": "Quattro asserzioni gia' verdi: core/ non importa Home Assistant; le entity platform non toccano la rete; core/ non importa moduli entity; const.py e' foglia. Meccanica del cricchetto: una regola si aggiunge nello stesso commit che la rende vera, nessuna si toglie.",
+      "reading_estimate_minutes": 9
+    }
+  ],
+  "excluded": [
+    {
+      "what": "Port del delta fork-only (20 file + 6 feature dentro file condivisi)",
+      "why": "Congelato fino alla risposta sulla governance del 31/08: portarlo su un master non protetto rischia di essere scavalcato o duplicato."
+    },
+    {
+      "what": "Fetta 1 (estrazione di transport/mqtt.py dal coordinator)",
+      "why": "Deliberatamente non iniziata: il modello va concordato prima, altrimenti la prima fetta diventa un fatto compiuto."
+    },
+    {
+      "what": "Correzione dei quattro invarianti violati",
+      "why": "Citati come motivazione, non risolti: ognuno richiede il posto dove l'invariante possa vivere, che arriva con le fette."
+    }
+  ],
+  "reading_estimate_minutes_total": 32
+}
--- a/.code_reader/deliveries/2026-08-19_process-contributing-agents.json
+++ b/.code_reader/deliveries/2026-08-19_process-contributing-agents.json
@@ -1,0 +1,77 @@
+{
+  "delivery_id": "2026-08-19_process-contributing-agents",
+  "date": "2026-08-19",
+  "repo": "chery-connect-ha/omoda9-ha",
+  "intent": "Pubblicare l'impianto di collaborazione come regola scritta, in risposta a due obiezioni reali: 'la mia approvazione sarebbe una firma, non una review' (Caslinovich, non sviluppatore) e il rischio che una regola uniforme separi un contributore dal proprio lavoro. Nessun codice runtime toccato: solo prosa e template.",
+  "baseline": "29fef39 (v1.13.0, canon/master)",
+  "head": "d274493",
+  "delivered_as": "PR chery-connect-ha/omoda9-ha#11",
+  "units": [
+    {
+      "block": "intent",
+      "path": "CONTRIBUTING.md",
+      "lines": 191,
+      "note": "La prima schermata ('In one page') e' il documento: le quattro cose tenute separate (strumento / protocollo / catalogo / integrazione) e i tre tempi di una modifica. Il resto e' dettaglio. Il pezzo che porta il peso e' la tabella del gate proporzionale al raggio d'azione, e la riga 'merge mai bloccato, stable si'.",
+      "reading_estimate_minutes": 11
+    },
+    {
+      "block": "contract",
+      "path": "AGENTS.md",
+      "lines": 226,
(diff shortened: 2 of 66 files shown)


drake69:
@JackRonan — three pull requests open at once against the same config flow, so here is the order I propose and why, plus one thing this one needs that the other two don't.

*Per the convention I set out on #7: the trial merges and the code checks below were run by the agent I drive, not read by me. They are all reproducible from the commands, and the decisions after them are mine.*

## They don't collide — and that is better than it sounds

I expected conflicts in `config_flow.py`, `strings.json` and the translations. There are none: all three pairs merge cleanly (`git merge-tree` against each pair's merge base, no conflict markers).

More usefully, **#16 and #17 compose by construction rather than by luck.** #17 puts the language dropdown inside `_region_fields()`, the shared helper, and this pull request's new `async_step_login_password` builds its schema with `**self._region_fields()`. So the language selector appears in the password flow on its own, with nobody having to remember to add it. That is the seam doing its job.

**What is not established is the three of them together.** Each is green alone; no CI run has ever seen the composed state, and "they merge cleanly" is a statement about text, not about a config flow that still works. That gap is what the `beta` label now exists to close.

## Proposed order

1. **#18 first.** Smallest, a pure routing fix, and it is the only one of the three that helps somebody *today*: it stops telling people their OTP was invalid when it was fine. That is the confusion @MassimoDe walked into on #5.
2. **#17 second.** Additive. One behaviour change worth naming in the release notes rather than leaving people to find it: new setups default to **English**, while existing entries without the field keep falling back to `it-IT`.
3. **#16 last**, and rebased **once** after the other two land — not after each.

Then label the merged state `beta` and walk all three login paths on a real instance before any of it goes stable. Composition is the thing to test, and it is the thing no green check has looked at.

## What this one needs that the others don't

#16 is the only pull request here that introduces a **credential**. New auth path, raw account password, 290 lines across 8 files including `core/session.py` and `core/prova_token.py`. Under the proportional gate in `CONTRIBUTING.md` that is squarely "shared code — one non-author read", and I would rather it got one than rode the one-week valve.

**I will do that read, and I will say plainly how.** I do not read diffs line by line unaided; I will go through it with the agent and then answer questions about it in my own words on this thread. If I cannot, the read did not happen and I will say so.

One thing already checked, offered as a starting point and not as the review: the filter that keeps the password out of the entry is there —

```python
self._data.update({k: v for k, v in user_input.items() if k != CONF_PASSWORD})
```

What I want to follow through is the rest of that promise: the reauth path, whether anything reaches the logs, and the subprocess environment.

Nothing here blocks you. If you would rather merge #18 and #17 yourself now, do — you are the author and the gate on those is merge-freely. Say the word if you would prefer I do it.

drake69:
## Review

I said on this thread I would do the non-author read and say how. Here it is, and how: I went through the diff with the agent I drive, file by file along the path the password takes. I did not read 290 lines unaided, and I am not going to pretend I did. Everything below is a line you can open, except one inference that is marked as such — and that one is the reason I am writing.

**What holds, checked rather than taken from the description.** The password is filtered out of the entry before anything is stored (`config_flow.py`, `self._data.update({k: v for k, v in user_input.items() if k != CONF_PASSWORD})`), and what persists is only the `login_method` marker that makes reauth ask for it again. It reaches the subprocess through the environment and never through argv (`session.py`, `env=_subenv(ctx, OMODA_PASSWORD=password)`), so it cannot be read out of `ps`. And `prova_token.call_password` prints `pwd_len=` rather than the value, with the response passed through `_redact`. Three claims, three lines, all true.

**One thing I would fix before this lands**, and it is not on any of those three paths.

`call_password` sends the password as a **query parameter**:

```python
r = requests.post(A.BFF + TOKEN_PATH, params=params, headers=H, timeout=20)
```

and that call is not inside a `try`/`except` — only the `r.json()` below it is. So on a network failure `requests` raises, and urllib3 formats that exception as *"Max retries exceeded with url: /auth/oauth2/token?username=…&password=…"* — the query string is inside the message. The traceback goes to stderr, and `session.login_with_password` does `out = (r.stdout or "") + (r.stderr or "")`, logs it at debug, and hands `_riga_utile(out, …)` back up to `config_flow`, which writes it with `_LOGGER.warning` and shows it to the user as the reason the login failed.

**So: if the backend is unreachable while somebody signs in with a password, the password lands in the Home Assistant log at warning level.** Not an exotic path — "the Chery backend is not answering" is a normal Tuesday here.

**Confidence, honestly:** this is read, not reproduced. It rests on how urllib3 formats `MaxRetryError`, and that is exactly the kind of thing that should be seen rather than deduced. You have the setup — point `bff` at something unroutable and try a password login, then look at the log. If the URL is not in the message, I am wrong and this whole section evaporates.

**And this may be the wrong fix anyway, which is the more interesting question.** Twenty lines above, the same file already says:

> `mobile → BODY (data=), email → QUERY (params=)`. È la differenza che fa coniare il…

`call_password` uses `params=` for both branches, including `loginType: mobile`. Two questions rather than two assertions: does password-plus-phone-number actually work, given the mobile OTP path needs the body? And if the backend accepts the body for email too, the credential leaves the URL entirely and the paragraph above stops mattering. That would be better than guarding the exception, because a password in a query string is also a password in **Chery's** access logs, which we cannot redact from here.

To be fair about it: the OTP path has used `params=` all along, so this is not a new sin. What changed is what is in the params. An OTP is single-use and dead in minutes; an account password is neither.

**Tests.** 290 lines of new authentication and no test file touched. I would not ask for coverage of the whole flow, but one test that asserts the password appears in neither the built params nor the printed output is about ten lines, and this project has a fresh, expensive reminder of why: the SMS bug on #14 survived for weeks precisely because the wrong field order still minted a valid token. Code that looks like it works is the failure mode here, and a test is the only thing that looks at it when nobody is.

**Verdict.** The design is right — mint once, live on the refresh token, never persist, and reauth that re-asks instead of pretending. The three security claims in the description are true as written. I would like the exception path closed, or shown not to exist, before this goes stable; the query-versus-body question first, because it may remove the problem instead of guarding it.

Nothing here needs to hold up #18 and #17.

*Per the convention on #7: the reading and the checks above were done by the agent I drive; the judgement about what to do with them is mine, and I will answer questions about it here in my own words.*

JackRonan:
Done — went with fix #1 (body), and it was the right call for the reason you suspected: the body handler *decrypts* the password, so the fix wasn't "just move it," it was "encrypt it correctly first."

The password isn't SM4 (that's the OTP code) — the direct-password path uses **AES-128-CBC/PKCS7, key=IV="w9R8Ag1KiL0pvMHc"** (the app constant in protocol-notes `RE_METHOD_AND_SNIFFING.md` §2.3). Verified live today:

- clear body → `IllegalBlockSizeException`
- SM4 body → `BadPaddingException` (right block size, wrong cipher)
- **AES-CBC body → token minted** ✅
- the same AES ciphertext in the *query* → rejected ("username/password wrong"), because the query path doesn't decrypt — which is also why only clear-query worked before.

So `7a2cb24` encrypts AES-CBC and sends it in the body. That closes the finding twice over: the credential is now **ciphertext** *and* **out of the URL** — gone from the HA log *and* from Chery's access logs, the part a query-side try/except couldn't reach.

Your test suggestion is in too: `tests/test_login_password.py` asserts the password is absent from the built params (with an AES round-trip), goes in the body not the query, never appears in the logged output, and that a network error exposes neither the password nor a traceback.

drake69:
Conflicts resolved and pushed to this branch — `master` merged in (including #14, which landed
just now and touches the same file), your commits untouched.

**What the conflicts were:** `config_flow.py` and `const.py`, both against the language selector
that reached `master` in #17. Both were additive collisions — two features adding their own block
at the same spot — so both sides are kept, nothing rewritten.

**One thing worth knowing, because it is the opposite of what the conflict looked like:**
`async_step_login_password` already builds its schema with `**self._region_fields()`, and #17 put
the language dropdown *inside* that helper. So the password login inherits the `Accept-Language`
selector for free — the OTP e-mail language setting applies to this path too, with no extra code.
Nothing had to be wired.

**Verification — done by the agent, not read line by line by me:** full suite on the merge result,
**653 passed** on Python 3.13; the three translation key sets still agree exactly (371 / 371 / 371).

**What is still not verified: the car.** Nobody here has run this path against a live account with
a password set. @kowi4 — you wrote on #27 that OTP codes stopped arriving to you even in the
official app and that login+password is the only method that works for you. This is the branch for
that. Once it merges I will cut a pre-release you can install.

PR #17, opened 2026-08-24
Add English/Italian language selector

### LINKED ISSUE #15: Add a language selector (English/Italian) to setup
Add a language dropdown on the integration's setup page (English/Italian to start; more can follow).

It would set the `Accept-Language` header (and any `lang`/locale request param) on the backend
calls, so OTP emails, SMS and server-returned messages come through in the chosen language instead
of always Italian.

- Default preserves current behaviour (no change for existing setups).
- Scope is small and self-contained: a config-flow option + threading the chosen locale into the
  request headers.
- Groundwork for broader localisation later.

JackRonan:
Closes #15

Adds a **language dropdown** to the setup form. It sets the `Accept-Language` header on the
backend calls, which is what decides the language of **OTP emails, SMS and server-returned
messages** — the code endpoints have no `lang` parameter, so the header is the lever.

### Behaviour
- New option in the setup form: **Language (OTP emails / SMS)** — English or Italian, easy to
  extend later.
- **New setups default to English** (`en-GB`).
- **Existing setups are unchanged:** entries without the field (and any internal call without a
  context) fall back to `it-IT`, exactly as today.

### Changes
- `const.py`: `CONF_LANGUAGE`, `LANGUAGES` (`en-GB`/`it-IT`), `DEFAULT_LANGUAGE`, `LANGUAGE_FALLBACK`.
- `core/context.py`: `CoreCtx.language` (+ `OMODA_LANGUAGE` in `ctx_da_environ`).
- `core/omoda_auth.py`: `headers_post` uses the context language.
- `core/login_omoda.py`: OTP-send headers (email **and** SMS) use the chosen language.
- `core/session.py`: `_subenv` forwards `OMODA_LANGUAGE` to the login subprocess.
- `coordinator.py`: reads `CONF_LANGUAGE` and threads it into the context.
- `config_flow.py`: the dropdown in the region fields.

Verified locally: `Accept-Language` follows the selected language; ctx-less calls and entries
without the field stay on `it-IT`.

### DIFF
--- a/custom_components/omoda9/config_flow.py
+++ b/custom_components/omoda9/config_flow.py
@@ -23,6 +23,10 @@ from homeassistant import config_entries
 from homeassistant.core import HomeAssistant, callback
 from homeassistant.data_entry_flow import AbortFlow
 from homeassistant.helpers.selector import (
+    SelectOptionDict,
+    SelectSelector,
+    SelectSelectorConfig,
+    SelectSelectorMode,
     TextSelector,
     TextSelectorConfig,
     TextSelectorType,
@@ -31,6 +35,7 @@ from homeassistant.helpers.selector import (
 from .const import (
     DOMAIN, CONF_EMAIL, CONF_PIN, CONF_VIN, CONF_TUSERID,
     CONF_PHONE, CONF_AREA_CODE, DEFAULT_AREA_CODE,
+    CONF_LANGUAGE, DEFAULT_LANGUAGE, LANGUAGES,
     CONF_BFF, CONF_TSP_HOST, CONF_CERTS_SRC, CONF_CHANNEL_ID,
     CONF_CAR_MQTT_HOST, CONF_CAR_MQTT_PORT, DEFAULTS,
     CONF_POLL_NORMAL, CONF_POLL_CHARGING,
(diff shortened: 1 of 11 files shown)


drake69:
CI has been red here since 03:49 and on #19 since 10:26, in case the notification got lost. Same cause on both, and it is one line each.

```
FAILED tests/test_capabilities_scheda.py::test_il_contesto_non_riceve_capability_inventate
AttributeError: 'C' object has no attribute 'language'      # here
AttributeError: 'C' object has no attribute 'tenant_code'   # #19
```

**The test is not broken — it is doing the job it was written for.** It builds a stub coordinator and runs the real `_build_ctx`, and its docstring says why it exists: an adversarial review showed that `CoreCtx` could fill its own capabilities and *the whole suite stayed green*. So it deliberately reads every attribute `_build_ctx` touches, which means adding a new context field makes it fail until you acknowledge the field.

**Fix:** add the new attribute to the stub in the same pull request that introduces it, next to `c.tsp_host, c.bff, c.channel_id = ...` around line 323 — `c.language = "en-GB"` here, `c.tenant_code = "…"` on #19.

Worth doing rather than working around: the failure is the test asking you to state that the context surface grew by one field. That is exactly what it is for.

*Per the convention on #7: the log reading and the trace back to the test were done by the agent I drive.*

GurliGebis (approved):
LGTM 👍

PR #18, opened 2026-08-24
Show "no vehicle" instead of "invalid OTP" when the code was valid

JackRonan:
When the OTP code is **correct** and the token mints, but the account has **no vehicle**, the
config flow was showing **"invalid OTP code"** — sending people to re-check a code that was fine.

The post-mint login check fails for a car-less account (no TSP user), `confirm_otp` reports
failure, and `async_step_otp` mapped *any* failure to `otp_invalid`. The most common way to hit
this is signing into the **wrong account** — e.g. logging in with the phone number when the car
is registered on the email login.

### Fix
- `async_step_otp` now routes on whether a token was **actually minted** (`_pending_token_minted`):
  if a token was written, it proceeds to vehicle discovery, which correctly reports **`no_vehicle`**;
  only a genuinely rejected code (no token) shows `otp_invalid`.
- The `no_vehicle` message now hints to try the other login: *"…this account has no vehicles. If
  your car is registered under a different login (e.g. your email address), sign in with that one."*
- Translations updated (en / it / strings).

Related to #5 — this is exactly the confusing failure a phone-only user hits when their car is on
a different (email) account. Independent of the SMS identity fix (#14); own branch off `master`.

### DIFF
--- a/custom_components/omoda9/config_flow.py
+++ b/custom_components/omoda9/config_flow.py
@@ -64,6 +64,22 @@ def _pending_token_path(hass: HomeAssistant) -> str:
     return hass.config.path(f"{DOMAIN}_pending_token.json")
 
 
+def _pending_token_minted(hass: HomeAssistant) -> bool:
+    """True se un token è stato davvero SCRITTO (OTP valido), anche quando `confirm_otp` ha poi
+    riportato un fallimento perché il login post-conio non è riuscito — di norma perché l'account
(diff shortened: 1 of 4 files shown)


JackRonan:
Folded into #14 — the no-vehicle message and the SMS identity fix are the same login-flow concern, so they now ship on one branch (`fix/sms-login-identity-order`).

PR #19, opened 2026-08-24
Add Omoda/Jaecoo vs Chery region presets

JackRonan:
Adds **brand/region presets** to setup, so a Chery EU account can be onboarded without hand-typing
gateway details.

Chery-International serves Omoda/Jaecoo and Chery from the **same** "legend" backend — the MQTT
broker, TSP console and mutual-TLS certificates are common per region; what differs per brand is the
**BFF gateway, TENANT-CODE, channelId and countryId**. This exposes those and offers one-click
presets:

- **Omoda / Jaecoo — Europe** (default, current behaviour): `legend-oj.omodaauto.nl`, tenant `300006`, channel `1`, country `1`.
- **Chery — Europe**: `eu-chery.cheryinternational.com`, tenant `300001`, channel `2`, country `2`.
- **Custom / other region**: leave every field manual (as today).

### Changes
- `const.py`: `CONF_TENANT_CODE` / `CONF_COUNTRY_ID` (+ `DEFAULTS`) and a `PRESETS` table.
- `config_flow.py`: a brand/region preset dropdown in the login forms; a preset fills BFF + tenant +
  channel + country, "Custom" leaves everything manual. Tenant/country now flow into the login context.
- `coordinator.py`: reads tenant/country from the entry and passes them into the context.

The Chery values were sourced from the official Chery EU app — the same `chery_legend` application
this integration derives from. Existing entries fall back to the historic Omoda values via `DEFAULTS`,
so nothing changes for current setups.

### DIFF
--- a/custom_components/omoda9/config_flow.py
+++ b/custom_components/omoda9/config_flow.py
@@ -23,6 +23,10 @@ from homeassistant import config_entries
 from homeassistant.core import HomeAssistant, callback
 from homeassistant.data_entry_flow import AbortFlow
 from homeassistant.helpers.selector import (
+    SelectOptionDict,
+    SelectSelector,
+    SelectSelectorConfig,
+    SelectSelectorMode,
     TextSelector,
     TextSelectorConfig,
     TextSelectorType,
@@ -33,6 +37,9 @@ from .const import (
     CONF_PHONE, CONF_AREA_CODE, DEFAULT_AREA_CODE,
     CONF_BFF, CONF_TSP_HOST, CONF_CERTS_SRC, CONF_CHANNEL_ID,
     CONF_CAR_MQTT_HOST, CONF_CAR_MQTT_PORT, DEFAULTS,
+    CONF_TENANT_CODE, CONF_COUNTRY_ID,
+    CONF_PRESET, PRESETS, DEFAULT_PRESET,
+    PRESET_OMODA_JAECOO_EU, PRESET_CHERY_EU, PRESET_CUSTOM,
     CONF_POLL_NORMAL, CONF_POLL_CHARGING,
     DEFAULT_POLL_NORMAL_MIN, DEFAULT_POLL_CHARGING_MIN,
     CONF_VEHICLE_NAME,
@@ -247,6 +254,10 @@ def _ctx_del_flow(hass: HomeAssistant, data: dict, token_path: str | None = None
         tsp_host=data.get(CONF_TSP_HOST, DEFAULTS[CONF_TSP_HOST]),
         bff=data.get(CONF_BFF, DEFAULTS[CONF_BFF]),
         channel_id=str(data.get(CONF_CHANNEL_ID, DEFAULTS[CONF_CHANNEL_ID])),
+        # tenant/countryId per-marchio: senza questi il login e la queryList partivano
+        # sempre col tenant Omoda (300006), quindi un account Chery non veniva riconosciuto.
+        tenant_code=str(data.get(CONF_TENANT_CODE, DEFAULTS[CONF_TENANT_CODE])),
+        country_id=str(data.get(CONF_COUNTRY_ID, DEFAULTS[CONF_COUNTRY_ID])),
     )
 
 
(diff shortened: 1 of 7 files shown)


drake69:
@JackRonan — this needs the same one-line bump you did on #14, and on the way to finding that out I broke something and fixed it. All three things below are worth knowing.

**1. @divisl's download does not work because the branch was renamed.** He reported it above: `feat/brand-region-presets` is now `feat/region-presets`, so the archive URL you gave him 404s. That is the manual route, and it is the route the `beta` label exists to remove.

**2. The label did nothing on this pull request, silently — and that is a real defect.** I applied `beta` here and to #14 at the same moment. #14 produced a pre-release; this one produced **no workflow run at all**, and no message saying why.

**I have not established why.** What I can say is what I checked: `beta.yml` is absent from *both* branches, so that is not the difference. The only difference I found is timing — `beta.yml` landed on `master` at 10:25, the head of `feat/region-presets` is from 10:24 and has not been pushed since, while #14 was pushed at 12:27 and ran. That is a correlation with one sample on each side, not a mechanism, and I am not going to dress it up as one. If anyone here knows GitHub Actions well enough to say, I would rather be told than keep guessing.

**What is certain is the failure mode, and it is the part that matters:** applying the label can do nothing at all and say nothing about it. `workflow_dispatch` does not depend on the pull request's branch — it runs from the default branch — which is why it is there as the way out.

**Practical rule until somebody works out the cause:** if you label a pull request and no run appears within a minute, the label did not take. Push the branch — merging `master` into it is the cheap way — and try again, or dispatch it manually.

**3. I cut a broken pre-release here, and deleted it.** Having found the label did nothing, I dispatched the workflow manually — **without checking the manifest**, which still says `1.13.0`. It produced `v1.13.0-beta.1`: valid semver, and it sorts *below* the `1.13.0` people already have, so HACS would have shown it as older than the installed version rather than as an update. It existed for four minutes and the release and tag are both gone.

I knew about that failure mode — I opened #22 this morning specifically to prevent it — and did it anyway by taking the manual route around my own guard. **#22 is now merged**, so the label path refuses that tag and prints the number to bump to instead.

**So: set `version` to `1.14.0` in `custom_components/omoda9/manifest.json`, commit here, and re-apply the label.** Same as #14. It will then cut cleanly, and @divisl gets an installable build of the Chery endpoints instead of a zip he has to extract by hand.

**@divisl** — nothing needed from you yet. When the pre-release exists you will be able to get it from HACS: the integration's ⋮ menu → *Redownload* → turn on **Show beta versions** → pick the `1.14.0-beta` entry → restart. Bear in mind what Jack said: with a Chery account you will need an e-mail address associated with it to sign in for now. And if the login fails, the useful thing to report is *where* it stopped, not that it stopped.

*Per the convention on #7: the checks behind this were run by the agent I drive. The mistake in point 3 was mine, and so is the decision to describe it here rather than quietly delete the tag.*

JackRonan:
Thanks Luigi — and for writing up point 3 rather than quietly deleting it; that's the more useful version.

Before I bump: **#14 is already on `1.14.0` and has cut `1.14.0-beta.1`.** If I set this branch to `1.14.0` as well, both open PRs are claiming the same version. Does the beta workflow just roll to `1.14.0-beta.2` here (i.e. the SMS fix and the presets are meant to land together in 1.14.0), or would you rather this ship as its own minor — `1.15.0`, beta `1.15.0-beta.1` — so the two betas don't sort into each other? Happy either way; tell me the number and I'll set it, merge `master` in first (per your "no run within a minute" rule, so the label actually fires), and re-apply `beta`.

On the rename: the 404 is on me — I cut `feat/region-presets` from a clean presets-only commit and deleted `feat/brand-region-presets`, which killed the archive link @divisl had. The beta build is the proper fix; once it cuts, that's the link he should use.

drake69:
Conflicts resolved and pushed to this branch — `master` merged in, now including #14 and #16 which
landed today and touch the same file. Your commits are untouched.

**What the conflicts were, and the one that needed a decision:**

- `config_flow.py`, in `_send_otp`'s context builder: additive — `tenant_code`/`country_id` from
  this branch alongside `language` from #17. Both kept.
- `tests/test_capabilities_scheda.py`: both sides taught the fake coordinator about a new
  attribute. Both kept.
- `_region_fields()`: **this one was not mechanical.** This branch and #17 each rewrote the
  docstring and the head of the returned dict. I merged them by hand rather than concatenating:
  the preset dropdown first (it is the normal choice and fills BFF + tenant + channelId +
  countryId), then the language dropdown, then the manual fields the presets leave alone. The
  docstring now says explicitly that **language is independent of preset** — it drives
  `Accept-Language`, not which region you connect to — because a reader landing on two dropdowns
  in one helper will otherwise assume the brand preset also sets the language.

**Verification — done by the agent, not read line by line by me:** full suite on the merge result,
**653 passed** on Python 3.13, and the three translation key sets agree exactly (381 / 381 / 381).

**Not verified: any car.** Neither the Chery preset nor the Omoda/Jaecoo one has been exercised
against a live account here — nobody in the group has a Chery-branded account. This stays
`unverified-hardware` on merge, and it is precisely the kind of change worth putting in front of
an owner as an installable build rather than describing.

PR #21, opened 2026-08-24
Bundle the Chery Lovelace vehicle card (auto-loaded)

### LINKED ISSUE #20: Port the bundled Lovelace vehicle card from omoda-jaecoo-ha
## Goal

Port the bundled Lovelace **vehicle summary card** from the Omoda/Jaecoo integration into `omoda9`, so a fresh install gets a curated dashboard card with no manual resource step — the same experience `omoda-jaecoo-ha` already ships.

## Source

From [`chery-connect-ha/omoda-jaecoo-ha`](https://github.com/chery-connect-ha/omoda-jaecoo-ha):

- `custom_components/omoda_jaecoo/lovelace/omoda-card.js` — the vehicle summary card (photo, name, battery %, range, charging state, warnings shown only when something's wrong; auto-discovers the integration's entities).
- `.github/workflows/card.yml` — the card's CI.

## Card name — DECIDED: `custom:chery-card`

Ship the element as **`chery-card`** (`customElements.define("chery-card", …)`), file `lovelace/chery-card.js`, picker label "Chery Card". Rationale: brand-level (Chery is the parent of Omoda/Jaecoo/Chery), collision-free against the existing `omoda-card`, forward-compatible with the issue #2 consolidation, and safe whether or not the proposed "Chery Connect" repo rename lands (it doesn't hard-code an unratified name). Since this is a brand-new card with zero installs, there's no dashboard-migration cost to choosing it now — but it IS a breaking change to rename post-release, so it's fixed here.

## What to change for omoda9

- **Discovery defaults**: the card defaults to `integration: "omoda_jaecoo"` / `prefix: "omoda_jaecoo_"`. Change these to `integration: "omoda9"` / `prefix: "omoda9_"` (our `DOMAIN` is `omoda9`). Both stay overridable via card config, as they are today.
- **Rename the element**: `omoda-card` → `chery-card` in `customElements.define(...)`, the JS class, and the file name (`chery-card.js`).
- **Auto-load**: replicate the source's registration in our `__init__.py` — `hass.http.async_register_static_paths([StaticPathConfig(...)])` serving the `lovelace/` dir, plus `frontend.add_extra_js_url(...)` with a cache-busting `?v=<version>`. Keep the URL path **integration-scoped**, `/omoda9_card/chery-card.js` (the path is tied to the `omoda9` domain, not the brand, so it can't collide with the Omoda/Jaecoo integration on a machine that has both). Guard it so it registers once, and swallow the failure path when the frontend isn't loaded (headless) — the card is still usable via a manual resource.
- **Header image**: wire the card's header photo to omoda9's image option if the integration exposes one; otherwise the card's `image:` option is the fallback.

## CI

Port `card.yml`: `node --check` on the card, **and** the guard that rejects optional chaining (`?.`) and nullish coalescing (`??`) — those fail to parse on older Android System WebView, which aborts `customElements.define()` and makes the card show "Configuration error" **on the companion app only**. The workflow strips comments before grepping (the file documents the rule in prose that mentions the tokens). Update the workflow's file path to `chery-card.js`.

## Setup instructions (both READMEs)

The repo is bilingual — add a **"Dashboard card"** section to **both** `README.md` and `README.it.md`:

- Auto-load: after a restart the card appears in **Add card → Custom → "Chery Card"**, no manual resource needed.
- Minimal config (`type: custom:chery-card`) and the documented options (`title`, `image`, `show_all`, `entities`, `integration`, `prefix`).
- **Companion-app fallback**: if the app shows "Configuration error", add the resource once manually — **Settings → Dashboards → ⋮ → Resources → + Add resource**, URL `/omoda9_card/chery-card.js`, type **JavaScript module** — then hard-restart the app.

## Acceptance criteria

- [ ] Card served + auto-loaded; appears in the card picker as "Chery Card" after a restart with zero manual steps.
- [ ] Registered as `custom:chery-card`; auto-discovers `omoda9` entities with an empty config.
- [ ] No `?.` / `??` in the shipped JS; card CI green.
- [ ] "Dashboard card" section in both READMEs, with the companion-app fallback.
- [ ] No collision when both `omoda9` and `omoda_jaecoo` are installed (distinct element name `chery-card` vs `omoda-card`, and distinct URL path `/omoda9_card/…`).

## Out of scope (follow-up)

The separate **energy dashboard card** (`www/omoda-jaecoo-energy-card.js` + `docs/energy-card-requirements.md`) is a distinct feature — tracked as a later issue, not part of this port.

Branch: `feat/lovelace-card` (draft PR against `master`, per #2).

JackRonan:
Closes #20.

Ports the vehicle summary card from `omoda-jaecoo-ha` into omoda9 and auto-loads it, so a fresh install gets a curated dashboard card with no manual resource step.

### What's here
- **`custom_components/omoda9/lovelace/chery-card.js`** — the curated summary (photo, battery %, range, charging, lock; warnings only when active). Registered as **`custom:chery-card`** — brand-level and distinct from the omoda-jaecoo `omoda-card`, so both integrations can coexist. Discovery defaults to `integration: "omoda9"` / `prefix: "omoda9_"`, both overridable.
- **`__init__.py`** — serves the card statically and `add_extra_js_url`s it (cache-busted by integration version). URL path is **integration-scoped** (`/omoda9_card/chery-card.js`) so it can't collide with another platform integration on the same machine; guarded to register **once per HA instance** (no-op from the second vehicle on); the frontend-missing (headless) path is swallowed so setup never fails on it.
- **`.github/workflows/card.yml`** — `node --check` + the CI guard that rejects `?.` / `??` (those fail to parse on older Android System WebView, which shows "Configuration error" on the app only). Comments are stripped before the grep.
- **`README.md` / `README.it.md`** — a "Dashboard card" section: minimal config, options table, and the companion-app manual-resource fallback.

### Verified locally
- `node --check chery-card.js` ✅
- WebView guard (no `?.` / `??`) ✅
- `__init__.py` compiles ✅

### Not in scope
The separate **energy dashboard card** (`omoda-jaecoo-energy-card.js` + its requirements doc) — a later issue, per #20.

Draft against `master`, per #2.

### DIFF
--- a/.github/workflows/card.yml
+++ b/.github/workflows/card.yml
@@ -1,0 +1,41 @@
+name: Card sanity
+
+on:
+  push:
+    branches: [master]
+    paths: ["custom_components/omoda9/lovelace/**"]
+  pull_request:
+    paths: ["custom_components/omoda9/lovelace/**"]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  card:
+    name: Lovelace card JS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+      - name: Syntax check
+        run: node --check custom_components/omoda9/lovelace/chery-card.js
+      - name: No optional chaining / nullish coalescing (older Android WebView)
+        # `?.` and `??` fail to PARSE on older Android System WebView, which aborts
+        # customElements.define() and makes the card show "Configuration error" on the
+        # mobile app only. Guard it in CI so a future edit can't silently reintroduce it.
+        # Comments are stripped first: the file documents this very rule in prose that
+        # mentions the tokens, which a naive grep would flag.
+        run: |
+          node -e '
+            const fs = require("fs");
+            const f = "custom_components/omoda9/lovelace/chery-card.js";
+            const code = fs.readFileSync(f, "utf8")
+              .replace(/\/\*[\s\S]*?\*\//g, "")
+              .replace(/^\s*\/\/.*$/gm, "");
+            const m = code.match(/\?\.[A-Za-z0-9_$([]|\?\?/);
+            if (m) { console.error("::error::" + f + ": optional chaining / nullish coalescing (" + m[0] + ") breaks older Android WebView"); process.exit(1); }
+            console.log("OK: no optional chaining / nullish coalescing in the card");
+          '
--- a/README.it.md
+++ b/README.it.md
@@ -116,6 +116,35 @@ l'app ufficiale dall'account. È una scelta tua, quindi parte **spento**:
   l'integrazione smette di interrogare il server: il rischio che l'account Chery
   venga bloccato cala molto, ma non sparisce → **correggi il PIN, non riprovare**.
 
+## Card per la dashboard
+
+L'integrazione include una card personalizzata e la **carica da sola** — dopo un riavvio
+compare in **Aggiungi card → Personalizzate → "Chery Card"** (nessuna risorsa da aggiungere a
+mano). È un riepilogo curato: nome del veicolo, batteria %, stato di ricarica, autonomia stimata
+e avvisi (gomme, batteria scarica, offline) mostrati **solo quando c'è qualcosa che non va**.
+
+La configurazione minima funziona già:
+
+```yaml
+type: custom:chery-card
+```
+
+Opzioni (tutte facoltative):
+
+| Opzione | Cosa fa |
+|---|---|
+| `title` | Titolo dell'intestazione (default: il nome del veicolo) |
+| `image` | URL dell'immagine di intestazione (ha la precedenza su quella nelle opzioni dell'integrazione) |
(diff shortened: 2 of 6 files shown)


drake69:
Hassfest is failing here, and it is a manifest line rather than anything about the card.

```
[ERROR] [DEPENDENCIES] Using component http but it's not in 'dependencies' or 'after_dependencies'
```

You import `StaticPathConfig` from `homeassistant.components.http` and call `hass.http.async_register_static_paths` to serve the card. Hassfest wants that declared.

**Fix:** add to `custom_components/omoda9/manifest.json`:

```json
"dependencies": ["http"],
```

The manifest currently declares neither `dependencies` nor `after_dependencies`, so this is a new key rather than an edit.

`dependencies` makes Home Assistant set `http` up first, which is what you want when serving a static path. `after_dependencies` would also satisfy hassfest and would fit your `if getattr(hass, "http", None) is None` guard — but `http` is always present in a real instance, so the guard is belt-and-braces either way and the plain dependency is the honest declaration of what the code needs.

*Per the convention on #7: the log reading and the check of what the manifest currently declares were done by the agent I drive.*

GurliGebis (approved):
LGTM 🙂

PR #24, opened 2026-08-24
docs: a security policy, and a private channel to use it with

drake69:
First piece of the fork-only port (#10, step 1), taken out of order because it removes a real gap rather than adding a feature.

**This repository had no way to receive a security report.** Private Vulnerability Reporting was **enabled on `omoda-jaecoo-ha`** — the line we have agreed to retire — and **disabled here**, on the repository that becomes the only one. So a project with a growing install base, which signs in to people's accounts and sends commands to their cars, could only be told about a vulnerability in public. I have enabled PVR on this repository; this file points at it.

I have a personal reason to care about this one. When I found the XSS in the Lovelace card in July there was nowhere to send it, and I had to decide on my own how long to sit on it. Nobody should have to make that decision about our code again.

**Ported, not copied.** The fork's `SECURITY.md` is good and most of it is here, but every factual claim was checked against *this* codebase. Two came back different:

- **The diagnostics section understated the code.** Redaction here also operates on values inside free text, not only on dictionary keys — some fields are exported verbatim — and it covers phone numbers, which matter as much as an e-mail address and, unlike an OTP, do not expire.
- **"Logs never contain the PIN, OTP, or tokens" is not something this codebase can promise.** Sign-in runs in a subprocess whose stdout *and stderr* are captured and logged; an unhandled error there produces a traceback we did not compose, so what a failure path carries is a property of the error rather than of our logging. The file states the normal case and then says exactly that, rather than an absolute that a single exception would make false.

**One addition of my own**: that Home Assistant stores config entries in plain text under `.storage/`. It is standard Home Assistant behaviour, it is out of our scope, and it is also where the credentials actually are. A security policy that leaves it out is decorative.

**On the rest of the port.** The fork-only delta is still entirely absent from this branch — verified item by item: none of the five CI workflows, no `NOTICE`, no `select.py`, no `core/otp_result.py`, neither Lovelace card, no `sandbox/`, and none of the charging-energy work. That is mine to bring across and it is queued. I am starting with the parts that touch no file @JackRonan currently has open in #14, #16, #17 and #18, so the port cannot collide with work in flight; `core/otp_result.py` waits, because it rewires error handling inside `config_flow.py`.

*Per the convention in `CONTRIBUTORS.md`: the checks against the codebase behind the two corrections above were run by the agent I drive; the decision about what this file may and may not promise is mine.*

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,79 @@
+# Security Policy
+
+## Reporting a vulnerability
+
+Report security issues **privately**. Please do not open a public issue, pull request or
+discussion for anything security-sensitive.
+
+- **GitHub → Security → "Report a vulnerability".** Private Vulnerability Reporting is
+  enabled on this repository, so the report is visible only to the maintainers until we
+  publish an advisory together.
+- Include the affected version, what the problem is, how to reproduce it, and what an
+  attacker gets out of it. A proposed fix is welcome and never required.
+- **Redact your own data before sending.** No real tokens, VINs, GPS coordinates, account
+  identifiers or phone numbers — the shape of a value tells us what we need
+  (`"vin": "<redacted, 17 chars>"`).
+
+You will get an acknowledgement as soon as one of us sees it. Coordinated disclosure is
+welcome: we will agree a timeline for the fix and the public advisory, and credit you by
+name unless you would rather stay anonymous.
+
+## Which versions get fixes
+
+This is a community project run by four people, none of whom does it for a living. Security
+fixes target the **latest stable release**. Update through HACS before reporting.
+
+Pre-releases exist here (HACS → *Show beta versions*) and are how a change gets tried on a
+real car before it lands. They are for testing, not for running on an instance you depend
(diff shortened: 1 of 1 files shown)


drake69:
**The channel has a first report waiting in it, and it is one of ours.**

@Caslinovich @JackRonan — there is a draft advisory in this repository's **Security → Advisories** tab. GitHub does not notify anyone about drafts, which is why I am saying so here; you are both repository admins so you can both see it.

It is **low severity** and concerns v1.13.0, not anything in flight. No detail here on purpose: sending it through the private channel and then explaining it in a public thread would make the policy in this pull request decorative on the day it was written. It is also not urgent — I would rather this channel got used for a small thing now than sat unused until something serious turns up and nobody has ever tried it.

One thing worth saying in the open, because it is about process rather than about the defect: the report includes what I have **not** done, which is reproduce it. It is derived from reading, and it names the one-minute experiment that would prove me wrong. A private channel is not a place where claims get to be softer.

*Per the convention in #23: the checks behind that report were run by the agent I drive; the decision to send it privately, and to say only this much in public, is mine.*

Issue #9, opened 2026-08-19
Tested-models matrix, as a capability report rather than a hand-kept table


drake69:
@JackRonan's idea from 4 August: rather than keeping a compatibility table by hand, have the integration ask the car what it is and build the table out of the answers.

**Why a hand-kept table would be wrong, not just tedious.** What a car accepts isn't a property of its model. Two cars of the same model on the same firmware can be refused different commands, and the same car has accepted a comfort macro in the afternoon and refused it that evening. A table saying "Jaecoo 7 cannot do X" is `if model ==` written in prose.

**So the unit is an observation, and it has four parts:**

```
verb:      windowControl
body:      {controlType: open, windowType: 2}   <- the one thing you varied
subject:   Omoda 9 - IT - fw <x> - declared: [...]
when:      2026-06-20 19:43:58
accepted:  A00079
acted:     fWinHeatingState 0 -> 1 at 19:44:12  <- empty if nobody watched
```

The two parts people leave out are the body and the date. The body, because the outcome depends on what you sent, not on the command's name — changing only `cycleData` turns `A00084` into `A00079`, and one denied field makes the backend refuse the whole request. The date, because an outcome is an observation and not a property.

**Refusals are as valuable as successes.** `A00084` and `A00079` are the rows that draw the boundary.

**The catalogue describes; it never decides.** It must not gate what we send: a stale row would take a working function away from someone who had it, which is what failing open forbids. At runtime, what the car declares wins. The rows are for people, for diagnosis, and for this matrix.

**Where the rows come from.** Two sources, same shape: the capability probe, and field-test reports from whoever owns the car (`docs/field-test.md`, and #11). The sandbox should be able to write a row out directly instead of printing it to the screen — it comes across with the port.

**And it's more than documentation.** The matrix says who owns which car, so it's also the register of who can check what — which is the same question as who can review what.

To start: say which car and region you have, and I'll seed it. @YasM0, @ThomasMeyer1970, @Sisku — you're the ones with models none of the four of us own, so you're the most useful people in this thread.

JackRonan:
Omoda E5, Region: UK.

Thinking how best to automate the capability test in a way that, should the users wish to, can be shared easily on here

YasM0:
Omoda E5 2026, UK. I believe Jack has the same model, possibly the same year?

PR #25, opened 2026-08-24
avvisi dei comandi: si leggono anche quando il comando viene rifiutato

Caslinovich:
## What changes, and why

> **This is older work, finished and never shipped.** It was written on
> **11 August 2026** and has been sitting uncommitted in a working tree since.
> It arrives as a pull request now, rather than as a push, because that is how
> a change lands here since `master` was protected. Nothing in it has been
> released to anybody: there is no tag and none is being asked for.

Warnings that explain what the integration skipped or adapted before sending —
a function the manufacturer does not authorise on that vehicle, a duration the
car does not accept — reached `sensor.omoda9_esito_comando` **only when the
command succeeded**.

A refused command raises inside `commands.send()`. The code that recomposes the
result with the warnings attached sat *outside* the `try`, so on that path it
never ran: the bare result of the last `emit` stayed on the state and the reason
went to the log alone. The warning was therefore missing in exactly the case
where a user needs it.

This is the defect described in [#1](https://github.com/chery-connect-ha/omoda9-ha/issues/1)
on 10 August — *"the notices go to the log only … the code that attaches the
notices to the result entity is never reached"*. It does not close that issue,
which has other open items.

Three changes, in `unisci_esito_e_note` and `_send_command`:

- **the result is recomposed inside the `finally`**, so it happens on the failure
  branch too. A failure to *display* the result must never replace the real
  error: it is caught and logged, because raising from there would replace the
  `CommandError` and with it the `reason` the coordinator routes the remedy on
  (PIN, re-authentication, permissions);
- **the command name is no longer repeated in front of every warning.** The
  result already opens with it, and those characters are needed inside the 255
  Home Assistant allows a state. Stripped only when the prefix is identical;
- **a warning that does not fit whole leaves the line** and adds to the count,
  instead of being chopped mid-word to make room for the `(+N in the log)` tail.
  A cut sentence looks complete, which is the one thing this function exists to
  prevent.

Relates to #1.

## Evidence

- [x] **Not testable by me** — labelled `unverified-hardware`.

Deliberately the most conservative row on the template, and it is worth saying
why rather than leaving it to be read as indifference:

- the display path is exercised by the suite, including a test that drives a
  real `A00084` refusal through the coordinator and asserts on `cmd_status`;
- **but nobody has watched a car do this.** On the author's Omoda 9 almost none
  of these warnings occur at all — they concern functions the manufacturer
  denies on *other* vehicles — so the author's own instance cannot falsify it.
  That is the ordinary condition here, not an excuse;
- the one case an Omoda 9 owner would see is a climate duration outside the
  accepted set followed by a refusal.

**So this is waiting for validation.** If you own a Chery-platform car and can
make a command be refused, what would settle it is one line: the text you see in
*Command result* after a refusal, and whether the reason is in it. No screenshot
— it would carry your plate or VIN.

## Checks

- [x] Suite green — 620 passed, gate R12 of `pr.sh`, on the exact commit pushed
- [x] No VIN, token, certificate, account id or raw capture — gate R02
      (`check_secrets.sh`, over the whole history, after the commit and before
      the push)
- [x] Fails open: nothing here can take away a function that worked for
      somebody. It only changes what is *written* on the result entity; the
      request sent to the car is untouched
- [x] Design docs updated if this contradicts them — it contradicts none

Five new tests in `tests/test_avvisi_visibili.py`. Three of them fail on the
previous code, so they are regression tests rather than decoration; the one that
matters is `test_il_comando_fallito_mostra_gli_avvisi_NELLO_STATO`, which
asserts on `cmd_status` under an `A00084`. The test that already existed
asserted on `_note_cmd`, which is why it caught nothing.

## Lettura richiesta

Touches shared code (`coordinator.py`): asking for one read by somebody other
than the author (R06). The merge is not blocked — if the week elapses with no
read, it is merged and labelled `merged-unread` (R20).


drake69:
## Review

Read by me with the agent I drive, along the path a refused command takes from `commands.send()` to the state of the result entity — per the convention on #7.

**The defect is real and the fix is in the right place.** The recomposition sat outside the `try`, so on the raising path it never ran; moving it into `finally` is what makes it happen on both branches. And catching inside that `finally` rather than letting it raise is the part that would be easy to get wrong: an exception there would replace the `CommandError` and take the `reason` with it, and `reason` is what the coordinator routes the remedy on. Losing it would turn a permissions problem into a re-authentication prompt.

**The test is the part I would point at.** `test_il_comando_fallito_mostra_gli_avvisi_NELLO_STATO` asserts on `cmd_status` under a real `A00084`; the test that existed asserted on `_note_cmd`, which is why it passed while the user saw nothing. That is the whole failure mode of this project in one line — a green check on the wrong surface — and it is worth more than the fix.

**Nothing here can take a working function away from anybody**: the request sent to the car is untouched. Under the proportional gate that is the lowest-risk shape a change to shared code can have.

---

**One thing that matters more than this pull request.**

You cite `pr.sh`, `check_secrets.sh` and gates `R02`, `R06`, `R12`, `R20`. **None of them exist in this repository.** So you have built the *executable* form of the rules that are written as prose in `CONTRIBUTING.md`, and you are the only person who can run it.

That is the same shape as the problem I spent this morning on: the review rule was unenforceable because a pull request produced nothing anyone could install. Here the standard is unreachable because the thing that enforces it lives on one machine. Both of your pull requests today are noticeably better documented than mine, and I do not think that is a difference in care — I think you have a gate and I do not.

**Please contribute it**, even rough, even if it only runs on your setup. A `tools/` directory and a paragraph in `CONTRIBUTING.md` mapping R-numbers to the prose rules would let @JackRonan and me hold the same line without having to remember it. If some of it is too tied to your machine, the interesting half is the checklist, not the shell.

**Practical note on trying this on a car.** Labelling this `beta` will now *fail* rather than publish: the manifest says `1.13.0`, which is the current stable, and a `v1.13.0-beta.N` tag sorts *below* the version people already have — HACS would show it as older. #22 landed an hour ago and stops it with a message naming the number to bump to. Set `version` to `1.14.0` and the label works. I found that out by publishing a broken one myself and deleting it; the details are on #19.

Waiting on a read from @JackRonan or on somebody making a command be refused. Not blocking the merge.

Caslinovich:
This needs a read, and I would rather ask than use the week.

*Per the convention on #7: the checks below, and the review that found the defect, were done by
the agent I drive. I have not read the diff myself.*

**The ask.** This touches `coordinator.py`, which `CONTRIBUTING.md` puts in the shared-code row:
CI green plus one read by somebody other than the author. I am the author, so I cannot supply it.
@drake69 @JackRonan — whoever has time. It is two functions, not one, and there are two places
worth attacking:

- the move of the result recomposition into `finally`, so it also runs when the command was
  refused — the behaviour this is named after;
- the rewritten truncation loop in `unisci_esito_e_note`, which decides what gets dropped when
  the warnings do not fit in the 255 characters Home Assistant allows a state. That is the part
  most likely to be wrong in a way the tests do not catch, and it is the one I would read first.

**One defect found and fixed while preparing this ask** (`a96f91c`), because it is more useful to
you than a clean-looking diff. The recomposition had been moved into the `finally` — correct —
but left *outside* the `try` that guards the publish. Only `_update` was protected. So if
`unisci_esito_e_note` raised — the very function this rewrites, and therefore the most exposed of
the two — the exception left the `finally` and **replaced the in-flight `CommandError`**, taking
the `reason` with it. That field is what the coordinator routes the remedy on, so a formatting
error would have reached the user as "reconfigure the vehicle-control PIN". The remedy is chosen
from the cause, and the cause belongs to the car.

It is now inside the `try`, `finale` starts from the bare outcome so there is always something to
return, and there is a test that makes the recomposition raise on a refused command and demands
that the `CommandError` and its `reason` survive. It fails on the previous commit. Suite: 616
passed, 5 skipped.

**The field test is not mine to give, and I am not going to pretend otherwise.** Almost every
warning this makes visible concerns functions the manufacturer denies on *other* vehicles; on my
Omoda 9 they do not fire, so my instance cannot falsify this. The `unverified-hardware` label
stays, and if this ships stable before somebody has run it, that has to be named in the release
notes rather than folded in quietly.

@ThomasMeyer1970 — the reliable way to produce a refusal on your J7 is the **driver seat heating**
switch: that whole permission category comes back denied on your vehicle, which is where the
`A00084` work started. If you send it and it is refused, the one line that settles this is **the
text you see in *Command result* afterwards, and whether the reason is in it**. No screenshot — it
would carry your plate or VIN.

Being straight about what it costs you, because it is not two taps: you are running @JackRonan's
build, so this would mean moving to this one. If that is too much for a single line of evidence,
say so and it waits for somebody already here. That is a perfectly good answer and I would rather
have it than a favour.

— Rino

Caslinovich:
@drake69 — thank you for the read, and my previous comment crossed it: you had already done it
38 minutes earlier and I asked for it anyway. That one is on me.

Two things follow from the timing. Your read is on `d9f3f49`; after it came **`a96f91c`**, which
fixes something your review would have caught in one more step — the recomposition was inside the
`finally` but *outside* the `try`, so only `_update` was protected and a raise from
`unisci_esito_e_note` itself would still have eaten the `CommandError` and its `reason`. It is in
the `try` now, with a test that makes the recomposition raise on a refused command and demands the
original error survive. CI is green on that commit. So there is a piece you have not seen, and it
is the piece you were circling.

**On the gate: yes, it exists, and yes, we will put it in the repo.**

You read it exactly right. It is `regole.sh` — 25 numbered rules with real blocking checks —
plus `pr.sh` for opening a pull request through them, `check_secrets.sh` over the whole history,
and, since this morning, `commento.sh`, which is why this message looks the way it does: a public
comment now has to carry an explanation in Italian for the person approving it, and it is checked
for secrets and for unattributed first-person claims *before* it can be posted. That last one
came from a real near-miss, not from a design session.

You are right about the shape of the problem, too. A rule that only one machine can enforce is a
rule the other three of us are guessing at, and I would rather be held to the same line than be
the only one who can see it.

One honest caveat before it lands: the scripts name paths on my machine, so they need a clean-up
pass before they are public — that is a review job, not a copy job, and it is the reason this
arrives as a pull request rather than as a paste. The half you called interesting, the checklist
mapping R-numbers to the prose in `CONTRIBUTING.md`, is the part that ports without change.

— Rino

ThomasMeyer1970:
@Caslinovich — thank you for pricing it honestly rather than asking for a favour. The
answer is no, and for the reason you named: one line of evidence isn't worth migrating a
setup with 103 entities whose history and automations key on the current entity ids.

But it shouldn't wait on me. @kowi4 has the same car — Jaecoo J7 SHS PHEV, EU — and is
already on the canonical line at v1.14.0-beta.1, so nothing to migrate. If 214 is denied on
his car the way it is on mine, the driver-seat-heating refusal you want is two taps for him.
He's also the one @drake69 asked about #26, so it could be one session on one car settling
both.

@kowi4 — if you do try it: press driver seat heating, let it be refused, then read what
"Command result" says afterwards and whether the reason is in the text.

I'll take it once the domain rename lands and everybody migrates once.

kowi4:
after click on cool everything I am getting this: "Raffredda tutto: HTTP 200 A00084 — funzione non autorizzata su questa auto 🚫 (non è il PIN) · not authorised on this vehicle (this is not your PIN) · saltate 4 funzioni non autorizzate su questa auto · 4 skipped (not authorised)"

checking the full command resutls
24 sierpnia 2026
nastąpiła zmiana stanu na "Raffredda tutto: HTTP 200 A00084 — funzione non autorizzata su questa auto 🚫 (non è il PIN) · not authorised on this vehicle (this is not your PIN) · saltate 4 funzioni non autorizzate su questa auto · 4 skipped (not authorised)"
17:35:32 - 12 sekund temu
nastąpiła zmiana stanu na "invio Raffredda tutto (taskId:cache)…"
17:35:32 - 12 sekund temu
nastąpiła zmiana stanu na "Raffredda tutto: il costruttore non autorizza affatto questa funzione su questa auto: nessun adattamento può farla passare"
17:35:32 - 12 sekund temu
nastąpiła zmiana stanu na "Raffredda tutto: non autorizzate su questa auto, salto aria sedile guida, aria sedile passeggero, aria sedile post. SX, aria sedile post. DX"
17:35:32 - 12 sekund temu



Not sure if this is what you want @Caslinovich @ThomasMeyer1970 @drake69

Caslinovich:
@kowi4 — thank you for the state you copied earlier. One thing first, so you do not spend
time on the wrong build: the version you are running, `1.14.0-beta.1`, is the SMS-login
branch and does **not** contain this pull request. What you read is today's behaviour. That
is useful as a baseline, but it is not this change.

This change is now installable on its own:

**`1.14.0-beta.2`** → https://github.com/chery-connect-ha/omoda9-ha/releases/tag/v1.14.0-beta.2

In HACS: ⋮ → *Redownload* → **Show beta versions** → `1.14.0-beta.2` → restart Home Assistant.

What would settle it, and it is the test @ThomasMeyer1970 asked you for: press driver seat
heating, let it be refused, then copy what **Command result** says afterwards. The question
is whether the reason is in the text of the entity state itself — not only in the log,
where nobody goes looking.

⚠️ These are separate branches, not a stack. `beta.2`, `beta.3` and `beta.4` each carry one
pull request, and none of them carries the SMS-login fix that is in `beta.1`. One at a time.
Your token and certificates live in `/config`, outside the integration folder, so switching
does not ask you to sign in again.

*Per the convention on #7: the check that `beta.1` does not contain these three changes was
done by the agent I drive. Choosing what to ask you for is mine.*

kowi4:
it seems it works on my side, driver seat heating is up and running but it also turned on a fan, then when I tried to turn off driver heating, the heating was disabled but fan was still on

nastąpiła zmiana stanu na "Sedile guida riscaldato: HTTP 200 A00079 — comando accettato ✅ · questa funzione non è autorizzata sulla sua via abituale su questo veicolo: inviata tramite il clima, che quindi si accende insieme a lei"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "Sedile guida riscaldato: HTTP 200 A00079 — comando accettato ✅ · questa funzione non è autorizzata sulla sua via abituale su questo veicolo: inviata tramite il clima, che quindi si accende insieme a lei"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "Sedile guida riscaldato: HTTP 200 A00079 — comando accettato ✅"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "Sedile guida riscaldato: HTTP 200 A00079 — comando accettato ✅"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "invio Sedile guida riscaldato (taskId:checkPassword)…"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "invio Sedile guida riscaldato (taskId:checkPassword)…"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "Sedile guida riscaldato: questa funzione non è autorizzata sulla sua via abituale su questo veicolo: inviata tramite il clima, che quindi si accende insieme a lei"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "Sedile guida riscaldato: questa funzione non è autorizzata sulla sua via abituale su questo veicolo: inviata tramite il clima, che quindi si accende insieme a lei"
23:29:11 - 1 minutę temu
nastąpiła zmiana stanu na "conio taskId (checkPassword)…"
23:29:10 - 1 minutę temu
nastąpiła zmiana stanu na "conio taskId (checkPassword)…"
23:29:10 - 1 minutę temu

drake69:
@Caslinovich — from my side #25, #26 and #32 are all ready and I am deliberately not merging them for you.

All three are green, all three have the read they asked for, and @kowi4 has run two of them on a Jaecoo J7 — the model they were written for and that neither of us owns. Under `CONTRIBUTING.md` there is nothing left blocking the merge.

Two reasons I am leaving the button to you rather than pressing it.

**The field evidence is layer one, not layer two.** @kowi4's report on #26 is `A00079`, the backend accepting the recomposed request. What has not been said is whether the seats actually ventilated — he was reading the command result, not standing at the car. And on this pull request his second run turned up something new: driver seat heating came on **and so did a fan**, and switching the heating off left the fan running. That is about the underlying behaviour rather than about your change, but you are the person who will know instantly whether it is a symptom or a separate bug.

**And @hdueo got a different answer on #26** — Jaecoo 7 PHEV, Italy, on `1.14.0-beta.4`: *"il costruttore non autorizza affatto questa funzione su questa auto: nessun adattamento può…"*. Two Jaecoo 7 PHEVs, two outcomes. Worth understanding before it ships stable, and it is exactly the kind of thing the model catalogue in #9 exists to record.

Merged instead, since they needed nothing from anybody: #34 and #36 (Dependabot, the first fruit of the `dependabot.yml` ported yesterday), and #38 — which adds the `Closes #N` line and the rule about keeping the three language files in step, with the test that enforces the second one.

#35 went into conflict behind #34 and has been asked to rebase itself. First time the rule from #38 has been exercised, about four minutes after it merged.

drake69:
**Reviewed, and I would approve it.** The part I went after is the one the comments say they are protecting, and it holds.

**A `return` inside `finally` would have eaten the `CommandError`, and it does not.** The new work sits inside the `finally`, but `return finale` stayed outside it, so on the refusal path the exception still propagates and the `reason` the coordinator routes on survives. That is exactly the failure the comment describes — the user being asked for a PIN because of a formatting error — and it is worth saying out loud that the diff avoids it structurally rather than by luck.

**The `while fuori:` loop terminates on every input.** Each pass either returns or pops one accepted warning while incrementing the count, so `accettate` strictly shrinks and the `if not accettate` branch is the floor. The growing tail (`(+9` → `(+10`) is re-measured each pass rather than assumed, which is the case that would otherwise overflow by one character.

**The prefix stripping only removes an exactly-matching prefix**, and `test_si_toglie_solo_il_prefisso_IDENTICO` is the test that keeps it that way. One residual: the prefix is whatever precedes the first `": "`, so an outcome like `Errore: …` would strip `Errore: ` from a warning that legitimately began that way. Cosmetic, and the warning survives — noting it, not asking for a change.

⚠️ **Blocking on the ordering, not on the code.** This bumps `manifest.json` to `1.14.0` and adds a `## v1.14.0 — 2026-08-24` section, and so do #32 and #26. #32 goes in first per #10. Please rebase and **fold this changelog entry into the existing v1.14.0 section, with no further version bump** — otherwise the second merge conflicts and the third conflicts twice. Nothing else here needs to change.

*Per the convention: the review was carried out by the agent I drive — it traced the exception path, the loop's termination and the prefix rule. The decision is mine.*

drake69:
@Caslinovich this now conflicts, and the reason is the one I flagged in the review: #32 merged first and took `1.14.0`, so `manifest.json` and the `## v1.14.0 — 2026-08-24` section collide here.

The code needs nothing. What it needs is your call on the notes, which is why I have not resolved it for you: **rebase, drop the second version bump, and fold this entry into the existing `v1.14.0` section** — or, if you would rather these ship as their own release, bump to `1.15.0` and say so. Merging your release notes into somebody else's section is an editorial decision about your text, not a mechanical fix, and I am not going to make it on your behalf.

One thing that changed in the last hour and affects the rebase: `master` now requires branches to be up to date before merging, and the **"Update branch" button is enabled**, so from here on a branch that has fallen behind can be brought current from the browser. This one still needs a real rebase rather than the button, because of the conflict.

drake69:
@Caslinovich conflict resolved and pushed to this branch, `master` merged in, your commits
untouched. You were away and it was a conflict we created, not you.

**What it was:** `CHANGELOG.md` only, and it was the fallout of the version rule agreed on
25 August (#46) plus everything that has landed since. Both sets of entries are kept,
master's first, in both languages. `manifest.json` now matches `master` exactly, so this
pull request no longer raises the version, which is what the rule asks.

**Verification:** full suite on the merge result, **666 passed** on Python 3.13.

**On merging it.** It carries a non-author read (mine, on 24 August, declared as
agent-assisted) and it is labelled `unverified-hardware`, so under the model in
`CONTRIBUTING.md` it can land. The proof on a real car happens on the pre-release, not
before the merge, and that is now stronger rather than weaker: pre-releases are cut from
`master` and contain everything merged so far, so this reaches a tester alongside the rest
instead of alone on a branch. #43 has the detail.

Two things from that August review still stand and neither blocks:

- your `unisci_esito_e_note` truncation loop is the part you asked to be read first, and it
  is the part I read; the defect you found and fixed yourself while preparing the ask
  (`a96f91c`) was the interesting one;
- your `pr.sh` / `check_secrets.sh` gate still does not exist in this repository. #37 landed
  `tools/` in the meantime, so the shape now exists for it. Still worth having yours.

PR #26, opened 2026-08-24
«Raffredda tutto»: la macro negata in blocco si ricompone sulla porta consentita

Caslinovich:
## What changes, and why

> **This is older work, finished and never shipped.** It was written on
> **11 August 2026** and has been sitting uncommitted in a working tree since.
> It arrives as a pull request now, rather than as a push, because that is how
> a change lands here since `master` was protected. Nothing in it has been
> released to anybody: there is no tag and none is being asked for.

On the vehicle in [#1](https://github.com/chery-connect-ha/omoda9-ha/issues/1)
(Jaecoo 7 PHEV, EU) category `210` is denied outright, so "cool everything" has
no door at all and field-level pruning has nothing to act on: the user got
`A00084` and there was no way round it.

Its **constituent parts**, though, are authorised on that same car under the
climate: `2041`/`2043` and the ventilation of the two front seats,
`2049`/`20410`. So the macro is rewritten as **one** `airControl` request
carrying the same fields. This is not a way around the denial — it is the door
that backend authorises, and it is the same thing the owner would get by
pressing climate and then each seat separately in the official app.

### `permessi.ricomponi()` — four guards, each for a different reason

1. **the manufacturer's macro is open** ⇒ nothing is touched. On an Omoda 9 this
   function is a literal no-op: when the official route works we never replace
   it with an imitation of our own;
2. **the destination door is closed too** ⇒ nothing is touched. We do not
   recompose towards a second denial; the real refusal is information;
3. **a field we cannot translate** ⇒ the whole thing is abandoned. Better the
   entire macro, which will be refused, than a truncated one sent silently. We
   never decide for the user to leave a function out without saying so;
4. **accessory fields are all carried over**, denied ones included — removing
   them, and naming *which*, is `pota`'s job immediately afterwards on the new
   endpoint. Deciding it here would mean writing it twice.

**Cooling only, deliberately.** `heatingControl` is not in `RICOMPONI`: among
its fields is `steerWheelHeatSwitch`, for which we do not know an entry under
`204` (the wheel has its own category, `208`, with its own working endpoint).
Recomposing heat would mean either sending a field we cannot vouch for, or
dropping it silently. And it helps no known vehicle: on both real permission
lists we have, `209` is allowed, and pruning already covers it. Guard 3 is what
stops a distracted extension — it is code, not just a comment.

### `commands._setpoint_da_posizione()`

Crossing the door, `temperature` **changes meaning**: in the macro it is the LO
*position* (outside the settable range by construction), on `airControl` it is a
setpoint, which has to be inside it.

`_limita_temperatura` alone is **not** enough, and it is an easy mistake to make
from its name: it falls back to LO..HI, and LO..HI contains LO, so on this path
the check always passed. With no vehicle spec it sent the `15.0` hard-coded for
an Omoda 9; on a car declaring only the extremes, its own LO. The measured body
the backend accepts carries `16.0`.

### For whoever reads it

This introduces the **first `from ..const import` inside `core/`** — the only one
in the sub-package. It does not break the "no Home Assistant in `core/`"
invariant (`const.py` has no imports at all), and `core/` is not importable
without Home Assistant today anyway, because of `omoda9/__init__.py` — checked,
identical before and after this change. It is still a new dependency towards the
parent package. If you would rather not have it, the two values can come through
`ctx.caps` instead. Flagging it rather than leaving it to be found.

Relates to #1. Does not close it: rear defrost (`232`) and the seat category
(`214`) are denied on that car and no adaptation reaches them.

## Evidence

- [x] **Not testable by me** — labelled `unverified-hardware`.

The label is about **this code**: it has never run on a car that needs it. On
the author's Omoda 9 the recomposition is a literal no-op, so the author's own
instance cannot falsify it. That is the ordinary condition here.

The changelog entry in this pull request does report a real-car observation, and
the two are not in conflict — it is worth being precise about who measured what:

- what was measured, by @ThomasMeyer1970 on his Jaecoo 7 PHEV on **11 August**
  ([comment](https://github.com/chery-connect-ha/omoda9-ha/issues/1#issuecomment-5255028085)),
  is the **request body**, sent **by hand from the sandbox**: `A00079`, and he
  went out to the car and both front seats were actually ventilating — layer two,
  not just the backend's verdict;
- what has **not** been observed is this integration composing that body and
  sending it. The suite asserts the composed body is byte-for-byte the one he
  sent, which is a statement about the code, not about his car;
- **switching off has not been tried by anybody.** It is the symmetrical twin of
  the same request. If a car refused it, the switch does not falsely report
  "off" — it stays as it was and the error is shown.

**So this is waiting for validation.** What would settle it is one field test
from someone with a car where "cool everything" is denied: press the button in
Home Assistant, then go and put a hand on the front seat cushions. The form is
in [`docs/field-test.md`](docs/field-test.md). No screenshot — it would carry a
plate or a VIN.

## Checks

- [x] Suite green — 630 passed, gate R12 of `pr.sh`, on the exact commit pushed
- [x] No VIN, token, certificate, account id or raw capture — gate R02
      (`check_secrets.sh`, over the whole history, after the commit and before
      the push)
- [x] Fails open: on a vehicle whose macro is authorised, `ricomponi` returns the
      command untouched. Where it is denied, the alternative is the `A00084` the
      user gets today. Nothing that worked for somebody can be taken away
- [x] Design docs updated if this contradicts them — it contradicts none; the
      `core/`→`const` note above is flagged for the reader instead

Fifteen new tests in `tests/test_permessi.py`, including exact equality with the
measured body and the "extremes only" row — which is the one that shows `16.0`
is read from the vehicle spec rather than hard-coded.

## Lettura richiesta

Touches shared code (`core/permessi.py`, `core/commands.py`): asking for one read
by somebody other than the author (R06). The merge is not blocked — if the week
elapses with no read, it is merged and labelled `merged-unread` (R20).


ThomasMeyer1970:
I'll run the field test — my car is the case this needs: 210 is denied on it, so
`ricomponi()` isn't a no-op here the way it is on an Omoda 9. Put the `beta` label on and
I'll pull the pre-release.

Useful for the port, and it makes this easy: the two lines don't share a domain. I'm on
`omoda_jaecoo`, this one is `omoda9`, so HACS puts the pre-release in its own folder and my
daily setup keeps running beside it — no entity ids lost, nothing to roll back.

One question before I install: does the Chery backend tolerate two concurrent sessions on
one account, or does a second login invalidate the first? If it does, running both side by
side isn't actually possible and I'd need to do this as a switch instead. @Caslinovich will
know that faster than I'd find out the hard way.

Once it's running, what I plan to report:

- press "cool everything" in HA, then go out and put a hand on both front seat cushions —
  the test asked for in `docs/field-test.md`;
- press off afterwards and see whether the ventilation actually stops. You note nobody has
  tried the symmetrical request, and I'll be standing at the car anyway;
- the setpoint that actually went out, from the log — you expect 16.0 read from the vehicle
  spec rather than the hard-coded 15.0, and that's checkable here;
- a regression check on the categories that are allowed on this car (204, 208, 209, 215),
  since "fails open" is a claim about not taking anything away, and mine is a car where
  something could be taken away.

No screenshots, as you say — plate and VIN.

drake69:
## Review

Read by me with the agent I drive — per the convention on #7 — concentrating on the four guards, the temperature crossing, and the new import you flagged.

**The `_limita_temperatura` catch is the best thing in here.** LO..HI contains LO, so the check passed on this path while doing nothing, and the function's name is exactly what would stop somebody looking twice. A guard that always succeeds is worse than no guard: it spends the reviewer's attention and returns nothing. That deserves to become a named invariant somewhere rather than living as a fixed line.

**On the new `from ..const import` inside `core/`.** You are right that it does not break the invariant — `const.py` imports nothing, so nothing about Home Assistant arrives through it, and the ratchet in #6 stays green on all four rules either way. I checked; you could not have, since #6 is not merged.

Still, I would rather have it through `ctx`. `core/context.py` already exists precisely as the seam through which the outside reaches `core/`, and everything else in there arrives that way. One import is not a problem; being the *first* one is how a second becomes easy to justify, and the reason `core/` can be exercised without Home Assistant is that the arrow only points one way. Your own suggestion, `ctx.caps`, is the one I would take — but this is a preference about direction, not a defect, and I am not going to hold a working change over it. Your call.

**The heating exclusion is right, and it is right because it is code.** `steerWheelHeatSwitch` has no entry we know of under `204`; recomposing it would mean either sending a field we cannot vouch for or dropping it silently, and both violate the rule that we never remove a function without saying so. Guard 3 stopping a distracted extension is the difference between a decision and a comment about a decision.

**On the evidence, which you split correctly.** What @ThomasMeyer1970 measured on 11 August is the *body* — sent by hand from the sandbox, `A00079`, and he went to the car and watched the seats ventilate. What has not been observed is this integration composing that body and sending it. The suite asserts byte-for-byte equality with what he sent, which is a statement about our code and not about his car. That distinction is the one this project keeps having to relearn, and you drew it before anyone asked.

---

**The problem with getting this validated, and it is not yours.**

The one person who can falsify this is @ThomasMeyer1970, because it is written for his car and is a no-op on yours. **He is running the fork line, `omoda-jaecoo-ha` v1.8.0** — not this code. It bit us this morning already: he measured his petrol range properly at the car and reported it against `sensor.omoda_jaecoo_range_gasoline`, an entity this line does not have, so a correct measurement could not answer the question it was taken for.

And it is not a one-click fix for him: the two lines have different HA domains (`omoda_jaecoo` versus `omoda9`), so switching renames every entity id he has, and running both at once is not safe — same account, and the sessions evict each other.

So this is likely to wait for the domain rename, when everybody migrates once instead of twice. I have asked everyone to declare what they are actually running on #27. Naming it here so that "unverified" on this pull request does not get read as nobody caring.

**Practical:** labelling this `beta` will now fail rather than publish — the manifest is at `1.13.0`, the current stable, and that tag would sort below what people already have. #22 stops it and names the number. Set `version` to `1.14.0` first. I learned that by publishing a broken one and deleting it; details on #19.

Not blocking. Waiting on a second read, and on a Jaecoo.

drake69:
@ThomasMeyer1970 — you are right and I overstated it. Two corrections and one thing that may save you the trouble entirely.

**On coexistence, you are correct and I was not.** I wrote that "running both at once is not safe". The *installation* side is exactly as you describe: the two lines have different HA domains, so HACS unpacks them into separate folders and your daily setup is untouched. Nothing to roll back, no entity ids lost. I collapsed two different questions into one and answered the wrong one.

**Your actual question has an answer, and it arrived in another thread an hour ago.** @Caslinovich, writing on #7 about why he has retired `ha_bridge.py`: *"two consumers on one cloud account fight over the session and produce OTPs."* So the backend does **not** tolerate two concurrent sessions on one account — the installations can coexist, the logins cannot. Running both side by side would have you generating OTP codes and evicting yourself, which is the failure @kowi4 is living through right now (see #27: he opened the official app and now no code reaches him at all, anywhere).

So it would have to be a switch after all — but read the next part before you do that.

**@kowi4 has your car and is already on the canonical.** From #27: Jaecoo J7 SHS (PHEV), EU, running `omoda9-ha` at `v1.14.0-beta.1`. That is the same model this pull request is written for, on the right codebase, with nothing to migrate.

@kowi4 — would you be willing? What is needed is on this pull request: press "cool everything", then go and put a hand on both front seat cushions. On a J7, category `210` is denied, so unlike on the author's car this code actually does something. It cannot take anything away — if the recomposition is refused, you get the error you would have got anyway.

That would leave @ThomasMeyer1970 free to stay where he is until the domain rename, and do the thing only he has queued: the three numbers @Caslinovich asked for on #7 — raw `cruiseRange`, the dashboard petrol range, and `rangeUnit` from the same frame. Two people, two jobs, neither of you migrating twice.

**If neither of you can, I will label it `beta` anyway** so it exists and waits. The pre-release mechanism is no longer theoretical: @kowi4 is running one, installed from HACS, which is the first end-to-end confirmation any of us has that it works.

*Per the convention on #7: the coexistence claim I got wrong came from me, not from a check — I asserted it from memory of the README's warning about the official app, which is about the backend session and not about installation. That is the second correction I owe today and both were the same shape: a true fact applied to the wrong question.*

ThomasMeyer1970:
@Caslinovich — the three numbers, from the integration's diagnostics dump (raw coordinator
state, not entity states). Jaecoo J7 SHS PHEV, EU:

    cruiseRange       = '393'
    mileageSurplus    = '633'
    rangeUnit         = 1
    dashboard petrol range, read at the car: 633 km

Your errata holds: `cruiseRange` is miles on this car too. What made it look otherwise is
that the fork I'm running reads `mileageSurplus` — a different field carrying the same
distance in km — and `cruiseRange` isn't visible from that line at all. Two correct
readings, two fields.

The structure in the frame explains it. Every distance comes twice, and the mile fields are
derived from the km ones by truncation:

    odometer          8510  ÷ 1.609344 = 5287.9  → odometerMile          5288
    pureElectricRange   90  ÷ 1.609344 =   55.9  → pureElectricRangeMile   56
    mileageSurplus     633  ÷ 1.609344 =  393.3  → cruiseRange            393

So the backend is metric-native and derives the imperial values. Which means the unit
doesn't need resolving per vehicle at all — read the km field and let Home Assistant
convert, and the guess disappears.

Two more things in the same frame:

- `mileageSurplusMile = '56'`. It should be 393. It carries `pureElectricRangeMile`'s value
  instead. Anything reading that field gets nonsense on this car.
- `electricRange = '5576'`, alongside `pureElectricRange = '90'` and
  `dynamicPureElectricRange = '90'`. 5576 isn't a plausible range in any unit I can think
  of. I don't know what the field is; flagging it rather than guessing.

A caution: `oilSurplus` reads '56' in this frame, the same value as two of the mile fields.
Coincidence as far as I can tell, but don't read anything into it.

`rangeUnit` is 1 here while `averageFuelUnit` is 2 and `tirePressureUnit` is 1 — so those
flags do take more than one value, whatever they mean.

Caslinovich:
@kowi4 — @ThomasMeyer1970 offered to run this one and then found he cannot: two concurrent
sessions on one Chery account evict each other, so he would have to switch his daily setup
rather than run it beside. He pointed at you, and @drake69 asked the same. Here it is,
installable:

**`1.14.0-beta.3`** → https://github.com/chery-connect-ha/omoda9-ha/releases/tag/v1.14.0-beta.3

In HACS: ⋮ → *Redownload* → **Show beta versions** → `1.14.0-beta.3` → restart Home Assistant.

What it does: on a car where the manufacturer refuses "cool everything" as a whole — which
is your car, as the `A00084` you posted shows — it rebuilds the request from the parts that
are allowed, instead of giving up on all of them.

**Here the state text is not enough.** The result line can say the command was accepted and
the car can still have done nothing. What settles it is: press *Cool everything*, then go
out to the car and put a hand on both front seat cushions — did the ventilation actually
start? Then press it off and say whether it actually stops; nobody has tried the off
direction yet. The time you did it helps, so it can be matched against the frame.

If the rear seats stay out on your car, that is expected and worth saying so out loud.

⚠️ These are separate branches, not a stack: `beta.2`, `beta.3` and `beta.4` each carry one
pull request, and none carries the SMS-login fix in `beta.1`. One at a time. Your token
lives in `/config`, so switching does not sign you out.

kowi4:
@Caslinovich after cool everything here is what I got

nastąpiła zmiana stanu na "Raffredda tutto: HTTP 200 A00079 — comando accettato ✅ · saltate 2 funzioni non autorizzate su questa auto · 2 skipped (not authorised) · Raffredda tutto: questa auto non autorizza il comando unico: ricomposto in un'unica richiesta al · (+1 nel registro)"
21:19:28 - 2 minuty temu
nastąpiła zmiana stanu na "Raffredda tutto: HTTP 200 A00079 — comando accettato ✅ · saltate 2 funzioni non autorizzate su questa auto · 2 skipped (not authorised)"
21:19:28 - 2 minuty temu
nastąpiła zmiana stanu na "invio Raffredda tutto (taskId:cache)…"
21:19:28 - 2 minuty temu
nastąpiła zmiana stanu na "Raffredda tutto: non autorizzate su questa auto, salto aria sedile post. SX, aria sedile post. DX"
21:19:28 - 2 minuty temu
nastąpiła zmiana stanu na "Raffredda tutto: questa auto non autorizza il comando unico: ricomposto in un'unica richiesta al clima"
21:19:28 - 2 minuty temu


Looks like AC was started with 16degrees and driver and passanger seat ventilation was turned on. There is no ventilation on the rear seats in this car.

After cool everything off this is what I got:

nastąpiła zmiana stanu na "Raffredda tutto OFF: HTTP 200 A00079 — comando accettato ✅ · saltate 2 funzioni non autorizzate su questa auto · 2 skipped (not authorised) · Raffredda tutto OFF: questa auto non autorizza il comando unico: ricomposto in un'unica richi · (+1 nel registro)"
21:27:15 - 12 sekund temu
nastąpiła zmiana stanu na "Raffredda tutto OFF: HTTP 200 A00079 — comando accettato ✅ · saltate 2 funzioni non autorizzate su questa auto · 2 skipped (not authorised)"
21:27:15 - 12 sekund temu
nastąpiła zmiana stanu na "invio Raffredda tutto OFF (taskId:cache)…"
21:27:14 - 12 sekund temu
nastąpiła zmiana stanu na "Raffredda tutto OFF: non autorizzate su questa auto, salto aria sedile post. SX, aria sedile post. DX"
21:27:14 - 12 sekund temu
nastąpiła zmiana stanu na "Raffredda tutto OFF: questa auto non autorizza il comando unico: ricomposto in un'unica richiesta al clima"
21:27:14 - 12 sekund temu

And seems like seats ventilation wasnt really turned off but AC was turned off.

hdueo:
jaecoo 7 phev italia
Versione: 1.14.0-beta.4
 
2026-08-25 07:53:19.466 INFO (SyncWorker_20) [custom_components.omoda9.coordinator] [cmd] conio taskId (checkPassword)…
2026-08-25 07:53:20.390 INFO (SyncWorker_20) [custom_components.omoda9.coordinator] [cmd] Raffredda tutto: non autorizzate su questa auto, salto aria sedile guida, aria sedile passeggero, aria sedile post. SX, aria sedile post. DX
2026-08-25 07:53:20.390 INFO (SyncWorker_20) [custom_components.omoda9.coordinator] [cmd] Raffredda tutto: il costruttore non autorizza affatto questa funzione su questa auto: nessun adattamento può farla passare
2026-08-25 07:53:20.391 INFO (SyncWorker_20) [custom_components.omoda9.coordinator] [cmd] invio Raffredda tutto (taskId:checkPassword)…
2026-08-25 07:53:20.561 INFO (SyncWorker_20) [custom_components.omoda9.coordinator] [cmd] Raffredda tutto: HTTP 200 A00084 — funzione non autorizzata su questa auto 🚫 (non è il PIN) · not authorised on this vehicle (this is not your PIN) · saltate 4 funzioni non autorizzate su questa auto · 4 skipped (not authorised)

drake69:
**Reviewed, and I would approve it.** This is the most delicate of the three and the docstrings carry their own argument, so I checked the two places where the reasoning could have been right and the code still wrong.

**`_setpoint_da_posizione` cannot be defeated by a half-declared card.** My first concern was `caps.get("clima_min", CLIMA_MIN_DEFAULT)` returning an explicit `None` — which would raise, be swallowed by the `except`, and let the LO position through unclamped, silently restoring the exact defect this function exists to remove. It cannot happen: `coordinator.py:1616` assigns `clima_min` and `clima_max` together, both already floats, or leaves both absent. The `.get` default is therefore the real fallback and not a decoy.

**The `basso > alto` guard is the right shape.** An incoherent card returns unchanged rather than inventing an interval, which is consistent with the rule stated everywhere else in `core/`.

**The call site keys on the original endpoint**, `c.get("endpoint") in permessi.RICOMPONI`, not on the rewritten one — which is what makes the recomposition path distinguishable from the routing path. That distinction is the whole point of the change and it is made in the one place where both bodies are in scope.

**On `heatingControl` being deliberately absent.** The reason given — `steerWheelHeatSwitch` has no known entry under 204, and recomposing the warm macro would mean either sending a field we cannot vouch for or dropping it silently — is the correct call, and the guard that cancels the recomposition on an untranslatable field means the rule is in the code and not only in the comment. That is the pattern #38 asked for, applied before it was asked.

⚠️ **Same ordering note as #25.** `manifest.json` → `1.14.0` and a third `## v1.14.0 — 2026-08-24` section. #32 lands first; please rebase and fold this entry into the existing section without bumping again.

*Per the convention: the review was carried out by the agent I drive — it checked the caps population, the guards and the call site. The decision is mine.*

drake69:
@Caslinovich this now conflicts, and the reason is the one I flagged in the review: #32 merged first and took `1.14.0`, so `manifest.json` and the `## v1.14.0 — 2026-08-24` section collide here.

The code needs nothing. What it needs is your call on the notes, which is why I have not resolved it for you: **rebase, drop the second version bump, and fold this entry into the existing `v1.14.0` section** — or, if you would rather these ship as their own release, bump to `1.15.0` and say so. Merging your release notes into somebody else's section is an editorial decision about your text, not a mechanical fix, and I am not going to make it on your behalf.

One thing that changed in the last hour and affects the rebase: `master` now requires branches to be up to date before merging, and the **"Update branch" button is enabled**, so from here on a branch that has fallen behind can be brought current from the browser. This one still needs a real rebase rather than the button, because of the conflict.

drake69:
@Caslinovich conflict resolved and pushed to this branch, `master` merged in, your commits
untouched. Same as #25: it was `CHANGELOG.md` only, fallout of the version rule agreed on
25 August plus everything that has landed since, not anything wrong here. Both sets of
entries kept, master's first, both languages. `manifest.json` matches `master`, so this no
longer raises the version.

**Verification:** full suite on the merge result, **675 passed** on Python 3.13.

**@ThomasMeyer1970 — the way you get this on your car has changed, and it is better for
you, but it is not what you were told in August.**

You offered to field-test this because 210 is denied on your car, which makes
`ricomponi()` a real operation there rather than the no-op it is on an Omoda 9. That offer
is still the only way this gets proven, and nothing has replaced it.

What changed is the delivery. The `beta` label that would have cut a pre-release from this
branch is being removed (#43), because HACS turned out to offer whoever has "show beta
versions" on **the most recently published pre-release, not the highest version**. Builds
cut from separate branches therefore hand testers a rising number with sideways content: on
24 August, beta.4 held one pull request and none of the three before it.

So pre-releases now come from `master` and contain everything merged. For you that means:
this merges, a beta is cut from `master`, and you install one build that has this **and**
the other work in it. One installation instead of four, and what you report is unambiguous.

Your question from August, still unanswered and still worth answering before you install:
**does the Chery backend tolerate two concurrent sessions on one account, or does the second
login evict the first?** @Caslinovich would know fastest. It decides whether you can run this
beside your existing `omoda_jaecoo` setup or have to switch.

**On merging.** Non-author read on 24 August, `unverified-hardware`, and the two points from
that review still stand without blocking: the `_limita_temperatura` guard that always passed
is fixed here, and the `from ..const import` inside `core/` remains a direction preference
rather than a defect, since `const.py` imports nothing.

ThomasMeyer1970:
I'll wait for the release cut from master, then switch over, then test. I prefer that order to four branch betas, so no objections to the change in #43.

On the session question: it has already been answered in this thread. On 24 August you corrected your own earlier claim, citing @Caslinovich on #7 — two consumers on one cloud account fight over the session and produce OTPs. Installations can coexist, logins cannot. So it has to be a switch rather than side by side, and the question doesn't need chasing again. For #27: I'm on the fork line, omoda_jaecoo.

Let me know when the build is available and I'll migrate.

Issue #27, opened 2026-08-24
Which repository is your install actually tracking? Please check and post it


drake69:
Please check which repository your Home Assistant is actually installing from, and post it here. It takes a minute and it is not a formality — it cost us a measurement this morning.

**Starting with mine, because it is the worst one.** My instance runs **`v1.8.0` of `chery-connect-ha/omoda-jaecoo-ha`** — the fork line, which has not moved since 2 August. HACS reports `latest_version: v1.8.0` and considers me **up to date**, because on that repository it is. I cannot see v1.13.0. I cannot install the `v1.14.0-beta.1` pre-release I published an hour ago. I have been reviewing, merging and releasing code I am not running.

**Why this matters, with today's example.** @ThomasMeyer1970 measured his petrol range at the car and reported it against `sensor.omoda_jaecoo_range_gasoline`. The canonical repository does not have that entity: it has `range_benzina` and a separate `range_combinato`, mapped from a different backend field. So a careful measurement, taken properly at the vehicle, **could not answer the question we asked**, through no fault of his. He spotted it himself and said so.

That is what two codebases in the field costs. Not disagreement — wasted work by the people whose work is hardest to get.

## The check

**Quickest:** find the update entity for the integration (`update.omoda_jaecoo_update` or `update.omoda9_update`) and look at its **`release_url`** attribute. The repository name is in the URL.

**Or in the interface:** HACS → the integration's page → the ⋮ menu shows the repository it tracks.

## What to post

- The repository your HACS tracks: `omoda9-ha` or `omoda-jaecoo-ha`
- The installed version
- One entity id you have, so we can tell the two entity-naming lines apart — for example whichever of `sensor.omoda_jaecoo_range_gasoline` / `sensor.omoda9_range_benzina` exists for you
- Car and region, if they are not already in `CONTRIBUTORS.md`

@Caslinovich @JackRonan @GurliGebis @ThomasMeyer1970 @Sisku @MassimoDe @divisl @kowi4 @YasM0 — and anybody else running this.

## What we will do about it

**No one has to migrate today, and nothing you have will break if you do nothing.** But two things follow from the answers:

1. **If you are testing for us, you want to be on `omoda9-ha`.** Everything discussed here for the last three weeks — the per-vehicle permission work, the SMS fix, the language selector, the password login — is there and not in the other line. A test run against the fork tells us about code we are no longer changing.
[...]

Issue #3, opened 2026-08-12
oilSurplus is a percentage on the Jaecoo 7 PHEV, not litres


ThomasMeyer1970:
Filing this on both repos since the code is identical in each.

On my **Jaecoo 7 PHEV** the fuel sensor (`oilSurplus`, shipped with a hardcoded `"L"`) reads a percentage, not litres. At a full tank it showed **100** — the car's tank holds 66 L.

The car's own numbers confirm it, all from one screenshot:

| value | reading |
|---|---|
| fuel remaining | 81 |
| fuel range | 919 km |
| average consumption | 5.8 L/100 km |

919 × 5.8/100 = **53.3 L** actually in the tank. And 81 % of 66 L = **53.5 L** — 0.3 % apart. Read as litres, 81 L would mean roughly 1,400 km of range; the car reports 919.

Happy to test anything on the J7 side.

JackRonan:
Apologies for closing the issue...I had two tabs open and closed the wrong one.

PR #32, opened 2026-08-24
sensor: work out the range unit from the car's own data, not from a constant

Caslinovich:
## What changes, and why

- sensor: work out the range unit from the car's own data, not from a constant

## Evidence

Non dichiarata: pull request aperta senza terminale.

## Checks

- [x] Suite green (`pytest tests/ -q`) — cancello R12 di `pr.sh`
- [x] Entity count unchanged — `tests/test_entity_count.py`, dentro la suite
- [x] No VIN, token, certificate, account id or raw capture — cancello R02 (`check_secrets.sh`)
- [ ] Fails open: nothing here can take away a function that worked for somebody
- [ ] Design docs updated if this contradicts them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ dell'integrazione: aggiorna da **HACS → Omoda 9 / Jaecoo → Aggiorna**.
 
 ## [Non rilasciato]
 
+## v1.14.0 — 2026-08-24
+
 ### 🇮🇹 Italiano
 
 - **«Autonomia benzina (miglia)» ora capisce da sola in che unità parla l'auto.** È un
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,32 @@ dell'integrazione: aggiorna da **HACS → Omoda 9 / Jaecoo → Aggiorna**.
 
 ## [Non rilasciato]
 
+### 🇮🇹 Italiano
+
+- **«Autonomia benzina (miglia)» ora capisce da sola in che unità parla l'auto.** È un
+  sensore di diagnostica, e finora dava per scontato che quel dato arrivasse in miglia —
+  cosa vera sulla Omoda 9, ma dedotta da due sole letture di una sola macchina. Su un
+  modello che lo mandasse in chilometri, Home Assistant avrebbe convertito un numero già
+  metrico e mostrato **un'autonomia più lunga di una volta e mezza**. Adesso il confronto
+  con l'autonomia benzina normale, che arriva sempre in chilometri, dice quale delle due
+  cose sta succedendo. Sulla Omoda 9 non cambia niente: il numero è identico a prima.
+- **Un dato incompleto non fa più comparire un numero sbagliato.** Se in una lettura manca
+  il valore di confronto, quel sensore tiene l'ultimo numero buono invece di mostrarne uno
+  che potrebbe essere falso — e che sarebbe rimasto lì fino alla lettura dopo.
+
+### 🇬🇧 English
+
+- **"Petrol range (miles)" now works out for itself which unit the car is speaking.** It is
+  a diagnostic sensor, and until now it assumed that reading arrived in miles — true on the
+  Omoda 9, but worked out from two readings of a single car. On a model sending kilometres
+  instead, Home Assistant would have converted an already-metric number and shown **a range
+  half again as long as the real one**. It now compares against the ordinary petrol range,
+  which always arrives in kilometres, and that comparison settles which case it is. On an
+  Omoda 9 nothing changes: the number is identical to before.
+- **An incomplete reading no longer produces a wrong number.** When the value needed for
+  that comparison is missing, the sensor keeps the last good number instead of showing one
+  that might be false — and that would have stayed on screen until the next reading.
+
 ## v1.13.0 — 2026-08-10
 
 ### 🇮🇹 Italiano
--- a/custom_components/omoda9/manifest.json
+++ b/custom_components/omoda9/manifest.json
@@ -14,5 +14,5 @@
     "numpy>=1.21",
     "pillow>=9.0"
   ],
-  "version": "1.13.0"
+  "version": "1.14.0"
 }
--- a/custom_components/omoda9/sensor.py
+++ b/custom_components/omoda9/sensor.py
@@ -157,6 +157,82 @@ def _frame_batteria_degradato(rt: dict) -> bool:
     return v == 0.0
 
 
+def _cruise_range_miglia(rt: dict):
+    """`cruiseRange` espresso in MIGLIA — o niente, se l'auto non permette di deciderlo.
+
+    ── Il fatto misurato ────────────────────────────────────────────────────────────────
+    Il centralino manda ALCUNE grandezze due volte, in due unità, come campi fratelli.
+    Frame reale di un'Omoda 9 PHEV, 2026-08-24:
+
+        mileageSurplus     209 km   ↔  cruiseRange            130   (rapporto 1,6077)
+        pureElectricRange  145 km   ↔  pureElectricRangeMile   90   (rapporto 1,6111)
+        lFrontTyreKpa      283      ↔  lFrontTyrePsi           41   (rapporto 6,90)
+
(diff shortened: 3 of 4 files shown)


drake69:
## Review

Read with the agent I drive, per the convention. Four things, and the third is the one that could close this today.

**You rejected your own hypothesis, with data, four hours after proposing it.** This morning on #7 you asked @ThomasMeyer1970 for `rangeUnit` and said that if it were `2` on a J7 and `1` on an Omoda 9 the field would be the answer. This pull request says instead that `rangeUnit` has been `1` in every sample kept, that the `2` belongs to `avgHkPowerUnit` — a different field — and that "1 and 2 mean km and miles" is **not measured**. That is the right call and it retires an idea I had already written into our backlog as the likely fix.

What replaces it is better because it is a measurement rather than a convention: some quantities arrive twice in two units as sibling fields, with the ratios to prove it — 1.6077, 1.6111, 6.90 — and *not all of them do*. Deciding on a ratio between two fields the car actually sent, rather than on a selector whose meaning nobody has confirmed, is the difference between reading the car and guessing its dialect.

**Keeping the declared unit at miles while normalising the value is the right trade, and worth defending in the comment as strongly as you defend the rest.** Changing a sensor's `unit_of_measurement` makes Home Assistant treat it as a different series: long-term statistics split or get discarded. A reviewer coming to this cold might "tidy" the declared unit into kilometres and silently cost every user their history. The comment says the unit is stable so history is not broken; it might be worth saying *that* explicitly rather than only *what* is stable.

**The measurement you are waiting for can arrive today, and not from @ThomasMeyer1970.** You wrote that the case closes with a reading on a J7. @kowi4 declared on #27 this afternoon: **Jaecoo J7 SHS (PHEV), EU, running `omoda9-ha` at `v1.14.0-beta.1`** — the same model, already on this line, nothing to migrate.

@kowi4 — if you are willing, three numbers from the same moment would settle a question that has been open since 12 August: the raw `cruiseRange` value, the petrol range your dashboard shows, and `mileageSurplus`. The raw values are in the diagnostics download (⋮ on the integration → *Download diagnostics*), which redacts VIN, position and identifiers before you see it.

**One case I do not think is covered, and I can test it myself.** When the anchor is missing, branch (1) returns `None` and the sensor keeps its last known value. You argue — correctly — that returning the raw value would resurrect the defect. But consider a **pure BEV**: no petrol, so `mileageSurplus` is absent. If such a car also sends `cruiseRange` (as electric range, in kilometres), the anchor never resolves and the sensor freezes on its last value forever, where before it showed a live number that was merely mislabelled.

Both behaviours are wrong and I think yours is the safer of the two — a frozen value is visibly stale, a live wrong one is not. But it should be a decision rather than a side effect, and it is checkable: **I have an Omoda 5 Premium BEV** and can say whether `cruiseRange` and `mileageSurplus` appear at all on a car with no engine. I will post it on #3 rather than here.

**Practical:** no conflict with #33 — checked with a trial merge against its head; the two touch `sensor.py` in different regions. Whichever lands first, the other needs nothing.

**On the `Evidence` block**, which you left as *"non dichiarata: pull request aperta senza terminale"* — that is worth more than a filled-in template. The rest of the body is explicit that this is a defence rather than a fix for an observed defect, that the identification of `cruiseRange` as the imperial twin is an inference from three samples on one car, and that a guard was written and removed because it was dead code that also froze the sensor. Nothing here is dressed up as more certain than it is.

*Per the convention: the reading of the diff, the trial merge against #33, and the cross-reference to @kowi4's declaration on #27 were done by the agent I drive. The BEV question, and the offer to answer it from my own car, are mine.*

drake69:
Answering my own question, measured rather than reasoned. I said I would put this on #3; it belongs here, because it is about this pull request's code path.

**On a pure BEV, `cruiseRange` does not arrive at all.** From my own car just now — Omoda 5 Premium BEV, Italy, on the fork line, but the telemetry is the same car and the same backend:

| entity | value |
|---|---|
| `range_combined_estimate` (= `cruiseRange`) | **unavailable** |
| `range_electric` | 378 km |
| `range_total` | 378 km — identical, so no petrol component |
| petrol range | no such entity |

So the case I raised does not occur: with no `cruiseRange` in the payload, `_cruise_range_miglia` returns at its first line and the sensor stays `unavailable`, which is what it already was. **No regression on a BEV**, and the `None`-when-no-anchor branch cannot freeze anything that was live.

That was the one objection I had. It is gone, and it took a car rather than an argument — which is the point of the rule.

**One thing this does not settle**, and it is worth keeping separate: it says nothing about a PHEV where `cruiseRange` is present. The J7 reading is still what closes that, and @kowi4 can produce it without migrating.

**Tick the fails-open box.** You left it blank, and on this evidence it holds: the worst the new path can do is return `None`, which leaves the last known value in place. Nothing that worked for somebody stops working — on a BEV the sensor was already unavailable, and on a PHEV the anchor is present by construction, since `mileageSurplus` is what makes it a PHEV.

*Per the convention: the reading of my instance's states was done by the agent I drive, through the REST API. The car is mine and it is a BEV, which is the part that mattered.*

kowi4:
@drake69

mileageSurplus": "728", official app says there is 728km I can do on petrol

cruiseRange": "452",

Caslinovich:
@kowi4 — your two numbers settled this on their own, and I would rather say so than have you
install something for nothing.

`mileageSurplus` 728 against `cruiseRange` 452 is a ratio of 1.61, so on your car
`cruiseRange` is already in miles, exactly as it is on an Omoda 9. The branch converts only
when that ratio is close to 1 — which would mean the field arrived in kilometres — so on
your J7 it takes the other path and changes nothing at all. That is the answer @drake69 was
missing for the PHEV case, and it came from your dump rather than from an install.

The pre-release exists if you want it:

**`1.14.0-beta.4`** → https://github.com/chery-connect-ha/omoda9-ha/releases/tag/v1.14.0-beta.4

But the only thing it can tell us on your car is that nothing broke: the diagnostic sensor
*Autonomia benzina (miglia)* should still read about 728 km, the same as your dashboard and
the same as before. If it ever showed 452 km or about 1170 km, that would be a regression
worth knowing — otherwise there is nothing here for you to test.

If you have spare time for one of these, #25 and #26 are the ones that need your car to act.

*Per the convention on #7: the reading of the branch behind this, and the ratio above, were
done by the agent I drive.*

kowi4:
I dont see any regression from my side with this version

drake69:
**Reviewed, and I would approve it.** Read against the code rather than the description, and the two things I went looking for are both fine.

**The BEV worry is void, and I checked instead of assuming.** `_cruise_range_miglia` returns `None` when `mileageSurplus` is missing or zero, which on a car without a petrol tank would mean a sensor stuck at its last known value forever. It cannot happen: `range_combinato` is in `_SENSORI_SOLO_TERMICO`, so on a confirmed BEV the entity is never created. Worth writing down because the reasoning lives two files away from the change.

**No division by zero on any path.** `mi <= 0` returns early, so `km / mi` is safe; the twin-pair guard is `if e_km and e_mi and …`, so a zero on either side skips the guard rather than dividing by it.

**One case the description flags as unchecked, and it is real.** "Fails open: nothing here can take away a function that worked for somebody" is the unticked box, and the case is a PHEV — or a car whose power type is not confirmed — that sends `cruiseRange` without `mileageSurplus`. It used to show a number; now it keeps the last known one and never updates. The docstring argues that trade explicitly and I agree with the argument: a possibly-false number that also poisons the restored value is worse than a stale one. I am recording it so that if somebody reports a frozen petrol-range sensor, this is the first place to look rather than a mystery.

**On the threshold tests.** The four boundary cases at 0.92 / 0.93 / 1.07 / 1.08 are what makes the window a decision instead of a magic number, and the note about four surviving mutants is why they exist. Right instinct.

⚠️ **The one thing to settle before merge is not in this diff.** This PR, #25 and #26 each bump `manifest.json` to `1.14.0` and each add their own `## v1.14.0 — 2026-08-24` section. Only one can have that number. Per the order in #10 this one lands first; the other two fold their entries into the same section with no further bump.

*Per the convention: the review was carried out by the agent I drive — it read the diff, checked `_SENSORI_SOLO_TERMICO`, and traced the division paths. The decision is mine.*

drake69 (approved):
Approved — findings in the review comment above.

PR #33, opened 2026-08-24
feat(sensor): home/away charging energy and the At Home sensor, ported from the fork line

### LINKED ISSUE #29: Feature: home vs away charging energy sensors + "At Home" presence (BEV)
## What

Three new location-aware entities, built entirely on data the integration already has (`chargingPower`, `chargeState`, GPS `position`):

- **`binary_sensor.at_home`** — ON when the car is inside the Home Assistant `home` zone (same zone logic as the device_tracker), `unknown` until the first GPS fix.
- **`sensor.home_charging_energy` / `sensor.away_charging_energy`** — lifetime kWh counters (`TOTAL_INCREASING`, ready for the Energy dashboard) that estimate energy charged at home vs anywhere else, by integrating charging power over time and splitting on the home zone.

## Why

Lets users see — and cost-split — home vs public charging directly in the Energy dashboard, without external template/Riemann helpers. `Home + Away ≈ total energy charged`.

## Design notes

- Trapezoidal integral of `chargingPower`, sampled by the existing ~120 s charging poll (`CHARGING_POLL_EVERY`). Intervals wider than a cap (`CHARGE_ENERGY_MAX_GAP`, 300 s) are **not** integrated, so a sleep gap can't invent energy; while the position is `unknown`, neither counter moves (energy is never attributed to the wrong zone).
- It's an **estimate** (a few % off a real meter), not a fiscal reading — clearly documented.
- **BEV-only** (`chargingPower` is BEV-only) and **opt-out** via a new options-flow toggle, so PHEV/unknown powertrains never get empty counters.
- Zone computed via `homeassistant.components.zone.async_active_zone`, in a single shared helper reused by all three entities (single source of truth, mirrors the device_tracker's own home/not_home).

## Status

Implemented and running on my instance (v1.7.0) on a branch in my fork (`drake69`). Passes `hassfest`. Happy to open a PR if you're interested — let me know if you'd want any changes to naming, the opt-out default, or the BEV gating.

drake69:
Closes #29.

Three entities this line never had: two lifetime kWh counters splitting charging energy by whether the car sat **inside** or **outside** the Home Assistant `home` zone, and the binary sensor that says which. They drop straight into the Energy dashboard as sources — home cost against public columns.

## Ported, not rewritten

Every hook the fork used has an equivalent here, which is why this is additive rather than a rewrite: `get_rt_field` → `_rt()`, the charge-state convention, `chargingPower`, `data["position"]`, `coordinator.poll_enabled`, and `Omoda9Entity.__init__` takes the same arguments as its counterpart. The two `sensor.py` files turn out to be structurally parallel — same class hierarchy, same names modulo the prefix — with `ChargedEnergy` the one class the fork had and this line did not.

**Names and entity ids follow this line's convention** (Italian, like every other sensor in the file) rather than the fork's. The domain rename will change the ids for everybody anyway, and consistency inside the file is what a reader needs.

**`car_zone()` goes in `entity.py` as one shared source of truth** for the binary sensor and both counters. If those ever diverged a user would see the car at home and the energy counted as away, with no way to tell which was right.

**Taken from the fork and not from the merged repository, deliberately.** That merge lost the `At Home` sensor while keeping the energy card that references it three times — noted when the delta checklist was extracted, and this is where it would have bitten.

## This ships a known defect on purpose

`CHARGE_ENERGY_MAX_GAP` refuses to integrate across intervals wider than 300 s, so a sleep gap cannot invent energy. The other side of that is real and measured: **on slow AC charges with sparse polling almost the whole session falls into those gaps and the counter underreports** — 0.53 kWh counted against ~8.2 kWh reconstructed from SoC, on the fork line.

It is ported unchanged, and the reason is that a port has to stay verifiable against its source. Mixing the fix in here would mean nobody could tell whether a difference from the fork was the port or the correction — and on a counter that feeds the Energy dashboard, that is exactly the distinction worth keeping. The behaviour is pinned by a test so nobody removes the guard thinking it is a typo, and the constant carries the measurement in a comment. The fix is a separate change with its own evidence.

## The number that moves, and who needs to know

**Entity count 107 → 110.** @Caslinovich — this is the figure you adopted as your acceptance criterion for the architecture slices ("615 tests green, 107 entities, zero `unavailable`"). It moves here, deliberately, and `tests/test_entity_count.py` is updated in the same commit with the reason written next to it. After this it is **110**. Flagging it rather than letting you find it against a slice.

## What was checked, and what was not

**Checked:** every symbol the ported class needs exists in this line, verified one by one; all files parse; the three translation keys are present in `strings.json`, `it.json` and `en.json` with matching counts (41 sensors, 27 binary sensors).

**Not checked: the suite has not been run.** Home Assistant needs Python 3.13+ and this machine is on 3.11 — the same reason `validate.yml` carries a comment saying CI is the only place the suite is a real gate. So CI is the gate here, and if it is red the fault is mine rather than the port's.

**Not verified on a car at all.** Both counters are BEV-only (`chargingPower` exists only there) and depend on a `home` zone being configured. Nobody has watched this accumulate on this line. It ran for weeks on the fork, which is evidence about the code and not about this integration — the ordinary condition here.

Six tests, and three of them fail on the naive form of this code rather than decorating it: the long-gap rejection, the unknown-zone abstention, and `chargeState` arriving as `'1.0'` instead of `'1'` — which on the fork silently kept the integral at zero for a while.

## Two items leave the port list

`tests/test_normalize_phone.py` and `tests/test_reconfigure_phone.py` are already here under Italian names (`test_normalizza_telefono.py`, `test_reconfigure_telefono.py`). Still outstanding from the fork-only delta: `select.py`, `sandbox/`, `core/otp_result.py` — the last of which waits, because it rewires error handling inside `config_flow.py` where @JackRonan has four pull requests open.

*Per the convention in `CONTRIBUTING.md`: the symbol-by-symbol comparison between the two lines, and the transcription of the class, were done by the agent I drive. The decision to port the defect unchanged rather than fix it in passing is mine, and so is the naming.*

### DIFF
--- a/custom_components/omoda9/binary_sensor.py
+++ b/custom_components/omoda9/binary_sensor.py
@@ -152,8 +152,12 @@ class Omoda9ACasa(Omoda9Entity, BinarySensorEntity):
     finche' non c'e' un fix utilizzabile.
 
     Comodo per automazioni e schede che vogliono il booleano senza interpretare lo stato del
-    device_tracker. Condivide `car_zone` con i contatori di energia casa/fuori: se i due
-    divergessero, l'utente vedrebbe l'auto a casa e l'energia contata come fuori."""
+    device_tracker: risponde con la zona piu' PICCOLA che contiene l'auto, quindi in un
+    garage disegnato dentro casa dice `off`. E' la convenzione di Home Assistant e resta.
+
+    I contatori di energia NON usano questa funzione ma `in_zona_casa`, che chiede il
+    contenimento in `zone.home`: le due domande sono diverse e tenerle diverse e' voluto —
+    la presenza dice dove sei, l'energia dice a chi va addebitata."""
 
     _attr_device_class = BinarySensorDeviceClass.PRESENCE
     _attr_icon = "mdi:home-map-marker"
--- a/custom_components/omoda9/binary_sensor.py
+++ b/custom_components/omoda9/binary_sensor.py
@@ -159,7 +159,7 @@ class Omoda9ACasa(Omoda9Entity, BinarySensorEntity):
     _attr_icon = "mdi:home-map-marker"
 
     def __init__(self, coord) -> None:
-        super().__init__(coord, "A casa", "at_home", entity_id_format=ENTITY_ID_FORMAT)
+        super().__init__(coord, "Omoda9 A casa", "at_home", entity_id_format=ENTITY_ID_FORMAT)
 
     @property
     def available(self) -> bool:
--- a/custom_components/omoda9/binary_sensor.py
+++ b/custom_components/omoda9/binary_sensor.py
@@ -21,7 +21,7 @@ from homeassistant.helpers.restore_state import RestoreEntity
 
 from .const import DOMAIN, FIELDS_AS_RICH_ENTITY
 from .coordinator import SENSORS
-from .entity import Omoda9Entity, field_on
+from .entity import Omoda9Entity, car_zone, field_on
 
 
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, add: AddEntitiesCallback) -> None:
@@ -34,6 +34,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, add: AddEnt
     ents.append(Omoda9Online(coord))
     ents.append(Omoda9Awake(coord))
     ents.append(Omoda9Session(coord))
+    ents.append(Omoda9ACasa(coord))
     # — avvisi dal canale realtime (Round B): gomme + batteria scarica —
     for suffix, name, field, dc in _RT_BINARIES:
         ents.append(Omoda9RealtimeBinary(coord, name, suffix, field, dc))
@@ -143,6 +144,33 @@ class Omoda9RealtimeBinary(_Omoda9RestoreBinary):
     def _live_is_on(self) -> bool | None:
         rt = self.coordinator.data.get("realtime") or {}
         return field_on(rt[self._field]) if self._field in rt else None
+
+
+class Omoda9ACasa(Omoda9Entity, BinarySensorEntity):
+    """Se l'auto si trova dentro la zona `home` di Home Assistant, dalla sua posizione GPS
+    viva — stessa logica di zona del device_tracker. ON = a casa, OFF = fuori, `unknown`
+    finche' non c'e' un fix utilizzabile.
+
+    Comodo per automazioni e schede che vogliono il booleano senza interpretare lo stato del
+    device_tracker. Condivide `car_zone` con i contatori di energia casa/fuori: se i due
+    divergessero, l'utente vedrebbe l'auto a casa e l'energia contata come fuori."""
+
+    _attr_device_class = BinarySensorDeviceClass.PRESENCE
+    _attr_icon = "mdi:home-map-marker"
+
+    def __init__(self, coord) -> None:
+        super().__init__(coord, "A casa", "at_home", entity_id_format=ENTITY_ID_FORMAT)
+
+    @property
+    def available(self) -> bool:
+        # Deriva dalla posizione: con l'aggiornamento automatico spento il fix GPS non viene
+        # rinfrescato e questo riporterebbe un casa/fuori potenzialmente vecchio.
+        return bool(self.coordinator.poll_enabled)
+
+    @property
+    def is_on(self) -> bool | None:
+        zona = car_zone(self.hass, self.coordinator.data.get("position"))
+        return None if zona is None else zona == "home"
 
 
 class Omoda9Awake(Omoda9Entity, BinarySensorEntity):
--- a/custom_components/omoda9/const.py
+++ b/custom_components/omoda9/const.py
@@ -251,6 +251,17 @@ HV_ON_POLL_MAX = 90     # cap di sicurezza al numero di letture ravvicinate (~90
 # in ricarica una lettura realtime dà subito stato_ricarica/corrente_hv/tempo_residuo aggiornati.
 CHARGING_POLL_EVERY = 120   # secondi tra due letture realtime mentre la spina è collegata (carica)
 CHARGING_POLL_MAX = 300     # cap di sicurezza (~10h: copre una carica AC completa con margine)
+# Intervallo massimo (secondi) fra due campioni di ricarica ancora integrati nei contatori
+# di energia casa/fuori. Durante una carica il poll realtime gira ogni CHARGING_POLL_EVERY
+# (120 s); un buco piu' largo significa che l'auto si e' addormentata o ha smesso di
+# riportare, quindi NON si integra attraverso di esso — altrimenti si inventerebbe energia.
+# 300 s = si tollera un poll perso, si rifiutano i buchi di sonno.
+#
+# ATTENZIONE, difetto noto e portato di proposito senza correggerlo: su cariche AC lente con
+# polling rado quasi tutta la sessione cade nei buchi e il contatore SOTTOSTIMA. Caso reale
+# misurato sulla linea fork: 0,53 kWh contati contro ~8,2 kWh ricavati dal SoC. Correggerlo
+# e' un cambiamento a se', per non confondere il porting con il fix.
+CHARGE_ENERGY_MAX_GAP = 300
 # MARCIA (battito di rilevamento): l'auto IN MOVIMENTO non manda push MQTT (verificato dal vivo
 # 2026-06-24: a vettura in marcia la sessione MQTT è connessa ma non arriva alcun 5A02 → motore/
 # velocità restavano fermi al giorno prima) e il poll periodico "sveglia+leggi" è ogni ~ora. Senza
--- a/custom_components/omoda9/entity.py
+++ b/custom_components/omoda9/entity.py
@@ -29,14 +29,22 @@ def _flt(v):
 
 
 def car_zone(hass, position: dict | None) -> str | None:
-    """`"home"` / `"away"` / `None` (ignoto) per una posizione GPS viva, usando la logica di
-    zona di Home Assistant (`async_active_zone`) — LA STESSA che mette il device_tracker
-    dell'auto su `home`/`not_home`. `None` quando non c'e' un fix utilizzabile, cosi' chi
-    chiama puo' astenersi invece di tirare a indovinare.
+    """`"home"` / `"away"` / `None` (ignoto), con la semantica del DEVICE_TRACKER.
 
-    Sorgente unica condivisa dal binary sensor "A casa" e dai contatori di energia di
-    ricarica casa/fuori: se i due divergessero, l'utente vedrebbe l'auto a casa e l'energia
-    contata come fuori, e non avrebbe modo di sapere quale dei due ha ragione.
+    Usa `async_active_zone`, cioe' la stessa funzione che mette il device_tracker dell'auto
+    su `home`/`not_home`. `None` quando non c'e' un fix utilizzabile, cosi' chi chiama puo'
+    astenersi invece di tirare a indovinare.
+
+    ⚠️ **`async_active_zone` restituisce la zona piu' PICCOLA che contiene il punto**, non
+    `zone.home`. Una zona non passiva disegnata dentro casa — un garage, un vialetto, cioe'
+    esattamente il posto dove sta una wallbox — vince su `zone.home`, e questa funzione
+    risponde `"away"` per un'auto che e' a casa. Misurato da @Caslinovich: con una
+    `zone.garage` da 20 m sulle stesse coordinate, una ricarica domestica finiva nel
+    contatore "fuori".
+
+    Per una entita' di PRESENZA quel comportamento e' la convenzione di Home Assistant e
+    resta. Per un CONTATORE DI ENERGIA e' sbagliato: la domanda li' non e' "in che zona
+    sei" ma "sei dentro casa", e la risposta e' `in_zona_casa()`.
     """
     # import locale: nessuna dipendenza dal componente zone al momento dell'import del modulo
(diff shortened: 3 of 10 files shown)


drake69:
Green now, after one red run that is worth recording rather than force-pushing away.

The entity-count test found **26 and 39** — the old numbers — because the three new entities were being created as `sensor.energia_ricarica_a_casa` instead of `sensor.omoda9_energia_ricarica_a_casa`. Every entity in this integration is constructed with the domain in its name (`f"Omoda9 {spec.name}"` in the realtime sensors), the object_id is slugified from that, and the translation key is the same string with the prefix stripped. I ported the class faithfully and then named it the way the *fork* names things.

Two things follow.

**The test earned its keep.** Without it this would have shipped as three entities with ids inconsistent with the other 107, and users would have had to live with that — an entity id is not something you fix later. @Caslinovich, this is the number you adopted as your criterion doing exactly what you adopted it for, on somebody else's change.

**And it is the argument for CI being the gate here.** I wrote in the description that the suite had not been run locally, because Home Assistant needs Python 3.13 and this machine is on 3.11, so if CI came back red the fault would be mine rather than the port's. It did, and it was.

Still unverified and unchanged by any of this: nobody has watched these counters accumulate on this line. Both are BEV-only and need a `home` zone configured. Fixed in `4f8d650`.

drake69:
@Caslinovich — asking you for the read, because this is shared code and my own rule says I do not get to skip that step.

It touches `sensor.py`, `entity.py` and `binary_sensor.py`, so under the proportional gate in `CONTRIBUTING.md` it needs one read by somebody who is not the author. The one-week valve does not apply — it opened today. I would rather have the read than the valve.

**What is worth your attention, in order:**

1. **The entity count moves 107 → 110.** That is the number you adopted as your acceptance criterion for the slices. It is deliberate and `tests/test_entity_count.py` is updated in the same commit, but you should be the one to say whether moving it here is fine.
2. **`car_zone()` in `entity.py`** is new shared surface: the binary sensor and both counters compute home/away through it, so if they ever disagreed a user would see the car at home and the energy counted as away. It uses `async_active_zone`, the same function behind the device_tracker.
3. **The trapezoidal step and its guard.** It ships the known undercount unchanged and on purpose — the constant carries the measurement (0.53 kWh counted against ~8.2 from SoC) and a test pins the behaviour so nobody removes the guard thinking it is a typo. If you think porting the defect is the wrong call, that is the thing to say now.

**What I already got wrong here**, since it is the kind of thing you would find: I named the three entities with the fork's convention, without the `Omoda9` prefix, so the object_ids came out unprefixed and `test_entity_count` did not see them. CI caught it — your test caught it, doing exactly what you adopted it for, on somebody else's change. Fixed in `4f8d650`.

**Not verified on any car.** Both counters are BEV-only and neither of us can exercise them: yours is a PHEV, mine is a BEV but stays on the fork line until the domain rename. @JackRonan's Omoda E5 is the only instrument in the group, whenever he is back on a machine.

No rush — I am asking rather than waiting for the week to expire, and if you would rather spend the time on #14 or your own two, say so and I will ask @JackRonan instead.

Caslinovich:
@drake69 — read done, and it did not need the week.

## The suite, and the entities on a bench

You have Python 3.11 and Home Assistant wants 3.13+, so the run your description leaves
open is one I can do. On `4f8d650`:

```
616 passed, 5 skipped in 49.32s     # this PR
610 passed, 5 skipped in 48.68s     # origin/master
```

Exactly your six tests, nothing else moved.

Then I took the entities themselves for a drive, because the six new tests build the
counter with `object.__new__` and never touch Home Assistant. Integration started, `zone`
set up, car parked on the home coordinates, 7 kW fed through the coordinator for two
minutes:

```
sensor.omoda9_energia_ricarica_a_casa      0.233   # 7 kW x 120 s = 0.2333 kWh
sensor.omoda9_energia_ricarica_fuori_casa  0.0
```

Trapezoid, restore plumbing, zone split, entity wiring: all correct. Nothing below is
doubt about whether the code runs.

## 1. The BEV-only classification of `chargingPower` does not rest on anything

I started out intending to write the opposite of this section. `chargingPower` is the
first entry of `_RT_SENSORI_BEV` in `sensor.py`, under a comment that reads as settled
fact:

> Campi che ESISTONO solo sulle BEV: la cattura live dei 91 campi del 2026-06-25 ha
> dimostrato che questa vettura NON li manda, quindi si creano soltanto a BEV confermata

Your two counters are appended above that branch, so they are built for every car, and
that is an inconsistency: either the field is BEV-only and the counters should follow it
into the branch, or it is not and the comment is wrong. I went to check which, and the
answer is the second one.

The capture that comment cites is `30_capture/diag_awake_20260625.json`. It was taken
with:

```
chargeState           '0'
chargeGunState        '0'
fastChargingGunStatus '0'
```

Car not charging, cable not even plugged in. `chargingPower` is a quantity that exists
only while current is flowing; it could not have appeared in that frame on any car, BEV
or PHEV. The capture is real evidence of something, but not of this.

The same file already knows this is how the channel behaves — twenty lines above, about a
sibling field:

> remainChargeTime: ASSENTE ad auto non in carica (comparirà sotto carica)

And I confirmed it is still true today on the PHEV, twice over. Its live realtime frame
does **not** contain `chargeState`, while `sensor.omoda9_stato_ricarica` has been holding
*Ricarica completata* since 15:52 — the value survives in the entity because the sensor
keeps the last known reading, not because the field is still being sent. Better still, I
read the frame twice three minutes apart, car parked and plugged in, nothing happening:

```
21:18   93 fields
21:21   86 fields
```

Seven fields came and went while the car sat still. A field count is not a census of what
the car can send, and "the 2026-06-25 capture had 91 fields and `chargingPower` was not
one of them" is a statement about one moment, not about the model.

I also searched the whole capture archive: there is **no frame of an Omoda 9 taken while
it was charging**. Nobody has ever looked.

So I am not asking you to move the counters into the BEV branch, which is where this
review was going to end up. Doing that would withhold a working feature from PHEV owners
on the strength of a photograph taken with the plug out. What I would ask instead:

- **label it `unverified-hardware`.** Not for the reason in your description — that the
  counters are BEV-only and neither of you can exercise them — but for the sharper one:
  whether they are BEV-only is itself unmeasured. That is exactly what the label is for.
- **put the measurement on the list rather than the conclusion.** One read-only frame from
  a plugged-in Omoda 9 during a charge settles it in both directions: if `chargingPower`
  is there, `_RT_SENSORI_BEV` has an entry too many and PHEV owners get the counters; if
  it is not, your two counters belong in the branch and the comment was right by luck.
- **and say so next to the number.** 107 → 110 is fine by me, but the comment you added to
  `TOTALE_ATTESO` should record that two of the three are counted on a profile where
  nobody has yet shown they can move. That is a different statement from the other
  entities in that total, and the next person to read the number will not guess it.

I have a watcher running on my side that polls Home Assistant — never the car — every two
minutes and saves the first realtime frame it sees while the state says charging, then
stops. The car charged today, so this should not take long. I will report the field list
either way.

## 2. These are the first entities here that can go `unavailable` by design

`grep -n "def available" custom_components/omoda9/*.py` returns nothing on master. All
three new entities gate availability on `coordinator.poll_enabled`, which the coordinator
documents as:

> interruttore "Aggiornamento automatico" (switch): SPENTO di default — il poll sveglia
> l'auto, quindi parte solo se l'utente lo accende esplicitamente.

`async_start_telemetry_poll()` returns immediately when that is False and nothing in
`async_setup_entry` sets it True, so on a fresh install it stays False until the switch
restores an `on` state that does not exist yet. Measured, integration started:

```
coordinator.poll_enabled = False
sensor.omoda9_energia_ricarica_a_casa       unavailable
sensor.omoda9_energia_ricarica_fuori_casa   unavailable
binary_sensor.omoda9_a_casa                 unavailable
```

The only three entities in `unavailable` out of 110.

I am not asking for a change — I think your reasoning holds for the binary sensor, and I
will take it for the counters too. But it needs writing down somewhere a person will find
it, because `unavailable` has meant *orphan* in this project: it is how leftovers from an
entity-type change are spotted, and "107 entities, zero unavailable" is a check somebody
runs by hand after every release. After this it stops being a clean signal. A clause next
to `TOTALE_ATTESO` would do it.

(Aside, pre-existing and not yours: `Omoda9PollingSwitch`'s docstring says "ON di
default", contradicting the coordinator. Nothing depended on which was right until three
entities started to.)

## 3. `car_zone()` — faithful to the tracker, one hole for the counters, and no test

The claim that it reproduces the device_tracker is correct, and I checked it against the
installed Home Assistant rather than from memory: `TrackerEntity.state` does exactly
`active_zone is None → not_home`, `entity_id == zone.ENTITY_ID_HOME → home`, else the zone
name. Your function is that function.

The hole is what `async_active_zone` means. From `zone/__init__.py`:

> The active zone is the **smallest containing zone**, using distance to the zone center
> as a tie breaker.

So a non-passive zone drawn inside home wins over `zone.home` — a garage, a driveway, a
courtyard, which is to say the place a wallbox lives. Same end-to-end setup as above, car
on the home coordinates at 7 kW, with a 20 m `zone.garage` on the same point:

```
a casa                        0.0
fuori casa                    0.233
binary_sensor.omoda9_a_casa   off
```

Home charging, booked as away, with the presence sensor agreeing. A passive zone leaves it
correct — I checked that too — so the trigger is specifically a normal zone smaller than
home and containing the car.

For a presence entity that is the HA convention and I would not touch it. For an energy
counter it is the one thing the docstring promises will not happen. `async_in_zones()`
returns the active zone *and* the full list of zones containing the point; asking whether
`zone.home` is in that list answers the containment question the counters actually have,
while the binary sensor keeps the tracker's answer. That would mean the two stop sharing
one function, which cuts against your reason for putting it in `entity.py` — so the honest
fix may be a sentence in the docstring instead. Either is fine by me; leaving it
undocumented is not.

And it is the one new shared surface with no direct test: all six tests monkeypatch it out
(`monkeypatch.setattr(S, "car_zone", ...)`). The lat/lon fallback, the `zone.home`
comparison and the no-zone branch are untested, and it is the function you asked to have
read.

## What I am not arguing with

The ported undercount I am not going to relitigate in a read: your reasoning about keeping
a port verifiable against its source stands on its own, the constant carries the
measurement and a test pins it, and none of that is what a second pair of eyes is for.
One thing for that comment though: the guard is not its only source. The step also refuses any interval
with a non-active endpoint, so the last partial interval of every session is dropped too.
Small next to the gap, same direction — worth a clause so the eventual fix does not stop
at `CHARGE_ENERGY_MAX_GAP` and declare victory.

`_in_ricarica` accepting `'1.0'` agrees with this line's own vocabulary: `CHARGE_STATE_MAP`
maps `"1"` to `In ricarica`. Consistent, not merely tolerant.

Naming, Italian, two entities rather than one with an attribute: all right by me.

The probes above are throwaway files, not part of this branch. Say the word and I will
open them as a small follow-up — the `car_zone` ones in particular fill the gap under
point 3, and the wallbox one would have caught it.

---

*Attribution, per `CONTRIBUTING.md`: the reading of this diff, the two runs of the suite,
the end-to-end probes quoted above, the inspection of the 2026-06-25 capture and the
checks against the installed Home Assistant source were performed by the agent I drive, on
my machine and against my environment. The conclusion in point 1 reverses the one I set
out to write, and the correction came from a question my own car prompted rather than from
the code.*

drake69:
@Caslinovich — all three of your points are in, and the first one was a real defect rather than a nit. Green again.

**The zone question is now two questions, because it always was two.** `car_zone()` keeps the device_tracker semantics — smallest containing zone, which is the Home Assistant convention and right for a presence entity — and `in_zona_casa()` asks containment in `zone.home`, which is what an energy counter actually wants. Somebody charging in their own garage is charging at home, and no zone a user draws tomorrow should silently move energy from one counter to the other.

They stop sharing one function, which cuts against the reason I gave for putting it in `entity.py`. **The reason was wrong.** Two questions that differ should differ in the code; sharing was hiding the difference rather than guaranteeing consistency. Both docstrings now say which is which and why, and yours is the measurement quoted in `car_zone`'s.

**`tests/test_zone_casa.py`** covers the logic directly — you were right that it was the one new shared surface with no test, and that all six counter tests monkeypatched away the thing whose read I had asked for. Your garage case is the test that carries the file:

```python
assert car_zone(zone_casa_e_garage, CASA) == "away"     # device_tracker semantics
assert in_zona_casa(zone_casa_e_garage, CASA) is True   # containment: it is at home
```

**`TOTALE_ATTESO` now records both things you said the number stopped saying**: that these are the first entities here that can be `unavailable` by design, so "N entities, zero unavailable" is no longer a clean way to spot orphans; and that two of the three are counted on a profile nobody has shown they can exercise. **`unverified-hardware` applied**, for your reason rather than mine — not that the counters are BEV-only and neither of us can test them, but that whether they are BEV-only is itself unmeasured.

**On `chargingPower`: you convinced me, and I would have merged the mistake.** I appended those counters above the `_RT_SENSORI_BEV` branch without ever asking what put that list together. That the classifying capture was taken with the car unplugged, where the field could not have appeared on any vehicle, is the kind of thing that survives for years because the comment reads like a finding. Your watcher is the right instrument and I will take whatever it reports.

**One thing of yours I have not done**, deliberately: the second source of undercount — the step also drops the last partial interval of every session, not just the long gaps. You are right and it belongs in that comment, but it changes what the constant's note claims, so it goes with the fix rather than into the port. Tracked.

**And a red run of my own, since it is the same discipline.** My first attempt at the zone tests failed: setting `hass.config.latitude` does not create `zone.home` — the zone component does, when it is set up. Three assertions were testing against a harness that had never built the thing. Fixed in `2edb87c`, and the fixtures now assert the zone exists before yielding, so the next person gets told which half is wrong.

Thank you for running the suite and the bench. That was the part I could not do, and it is why the read was worth asking for rather than waiting out the week.

GurliGebis (approved):

PR #37, opened 2026-08-24
tools: the rules of this repository, in a form that can be run

Caslinovich:
## What changes, and why

`tools/` — the rules of this repository written so a machine can check them, which until
today existed only on my machine. @drake69 asked for it on #25 and the argument was right:
a rule only one machine can enforce is a rule the other three of you are guessing at.

- `tools/rules.sh` — 26 numbered rules, each with its source and its severity, plus the
  gates. `./tools/rules.sh list` prints the table; `check` is a read-only preflight.
- `tools/pr.sh` — lands a change: branch, gates in order, push, pull request filled from
  the repository's own template. The order matters — the secret gate runs *after* the
  commit and *before* the push, because against staged-only changes it sees nothing.
- `tools/release.sh` — `status` · `prepare` · `publish` · `abort`.
- `tools/check-secrets.sh` — every confidential pattern, over the **whole history** and
  the working tree, including files git has never seen.
- `tools/README.md` — the table @drake69 called the interesting half: every `R` number
  against the paragraph in `CONTRIBUTING.md` or `AGENTS.md` it comes from.

**None of it is required.** `git push` and `gh pr create` are in `AGENTS.md` and always
will be. These just refuse to let you forget something.

## What I took out before publishing it

This arrived as a pull request rather than a paste because it needed a review pass, and
the review found things worth naming:

- **My own e-mail address was hard-coded in `check-secrets.sh`**, in the line meant to
  stop it being published — and the pattern does not match itself, so the gate would have
  published it and said PASS. Anything personal now lives in `tools/private-patterns.txt`,
  which is gitignored, with a committed example.
- **`/root/…`, `/media/…`, Proxmox and VM commands** were written into two of the scripts.
  Once tracked they also made the info-disclosure check report *itself*, five hits, every
  run — a warning that always fires is a warning people stop reading. The patterns are now
  written so they cannot match themselves.
- **Everything assumed the scripts sat at the top of the tree.** Moved into `tools/` they
  reported that `core/` had ceased to exist. The root is resolved through git now.
- **The rule that kept this tooling out of the repository (`R14`) blocked the pull request
  that publishes it.** Verified by running it. Rewritten: private material stays out, the
  tooling does not.
- **GNU-only constructs** — `grep -oP`, `sed -i`, `base64 -w0`, `date -Is` — are gone.
  `base64 -w0` was in the push path, so on macOS this could not push at all.
- **The suite gate hard-failed when no local interpreter had pytest**, which is stricter
  than `AGENTS.md`, and that file explicitly allows skipping it and letting CI be the gate.
  It now warns, and the pull request body says the suite did not run locally.
- **Forks**: `pr.sh` now sends `owner:branch` as the head when `origin` is not this
  repository.
- **Every pull request opened without a terminal was labelled `unverified-hardware`**,
  including one that changes no runtime behaviour — a false line in the next release
  notes. `--evidence backend|not-mine|none` states it instead. It cannot say *verified on
  hardware*: that answer means somebody watched a car.

One more was found by this very pull request, on the first real run: `pr.sh` treats a
different `origin` slug as a fork and sends `owner:branch` as the head. But **a renamed or
transferred repository is not a fork** — this one moved to the organisation and my remote
still carried the old owner, so GitHub answered `422 head invalid` with the branch already
pushed and nothing to show for it. The API resolves the old name through a 301, so it now
asks before deciding.

And once that was fixed it produced the wrong title: with `--continue` and no message the
title came from `git log -1`, so the two-line follow-up above became the name of the whole
pull request. It takes the first commit beyond the base now. Those are commits here, left visible rather
than folded away: a tool for opening pull requests that had never opened one is worth
knowing about, and so is the fact that CI caught what my own review did not.

And CI found the one the audit missed. `R14` — *the tooling stays out of the repository* —
is written down **twice**: in the shell gate, which I had rewritten, and in
`tests/test_repo_hygiene.py`, which I had not. It matches by basename, so `tools/release.sh`
failed the very pull request that publishes the tooling. Those names are now forbidden only
at the **root**, which is where the stale private copies actually sit.

That test had also been skipping itself: it asks whether `.git` is a *directory*, and inside
a git worktree `.git` is a *file* pointing at the common directory. Five tests had quietly
not been running for anyone who works in a worktree — which is why this passed locally as
`610 passed, 5 skipped` and failed in CI. It is `615 passed, 0 skipped` now.

## What it found on its first honest run

Two machine paths of mine already committed in `core/commands.py` and `tests/fixtures.py`.
Comments only, and they are scrubbed here. Two more (`/root/.local/…` in `config_flow.py`
and `tls_client.py`) are *correct* — that is where pip puts a package inside the Home
Assistant container — so the checker has one explicit allowance for them.

Both strings stay in the published history, where they cannot be removed without rewriting
it. Scrubbing the current text stops them spreading; it does not undo the past, and I would
rather say that than imply otherwise.

## Two things to argue with

**`R04` (a human, bilingual changelog) and `R05` (one stable release per fortnight) are
house rules with blocking gates.** They are my practice, not a decision anybody agreed to,
and if you cut a release they will stop you. They are marked **house rule** in the table
rather than dressed up as project rules. If the group wants them they belong in
`CONTRIBUTING.md` as prose — a separate pull request — and if it does not, delete the
gates. Same for `R15`, `R17` and `R18`, which have no gate at all.

**`R25` asks, it does not verify.** No program knows who really read a file. It recognises
explicit first person and, when the text does not say whose work it is, makes you answer
and writes the answer at the bottom. Calling it verification would be the exact thing this
project forbids.

## What is deliberately not here

The comment tool and the deploy script. The first exists because I do not read code and
should not approve English text blind, so it demands a summary in Italian first — that is
one person's problem, not the project's. The second copies the component into one
particular VM. `R26` belongs to the comment tool, so it is listed with no gate rather than
quietly dropped: a hole in the numbering would confuse more than an honest line.

## About the gate itself

A derogation needs a person at a terminal, a one-time code printed at that moment, and a
written reason, and it is refused if the register cannot be written. **It is a high kerb,
not a wall**: `script -qec` or `expect` can read the code and type it back, and that is
said at the top of the file rather than left for somebody to discover. There is a second,
narrower door — `--authorized "<what you were told>"` — because most of us drive an agent
and an agent has no terminal; it opens only for NORMAL-severity rules, and it records the
sentence word for word. It cannot verify that sentence, and it says so.

## Evidence

**No runtime behaviour** — tooling, documentation and two comment lines. Nothing a user
can see changes.

> *Attribution (R25): the audit, the rewrite and the testing in this pull request were done
> by the agent I drive, on my machine, and I have not read the diff line by line myself.
> The decisions — what to publish, what to keep private, which rules to mark as mine
> rather than the project's — are mine.*

## Checks

- [x] Suite green — `615 passed, 0 skipped`, run by the R12 gate itself
- [x] No VIN, token, certificate, account id or raw capture — `tools/check-secrets.sh`
      over the whole history: `RESULT: PASS`
- [x] Fails open: nothing here can take away a function that worked for somebody — no
      runtime code is touched
- [ ] Design docs updated if this contradicts them — n/a

Also exercised in a sandbox with a fake remote, never against this repository: the full
`pr.sh` run to the point of opening the pull request, `release.sh status`, the version bump
and changelog dating, the archive builder, and both derogation paths (a NORMAL rule waived
and recorded, a HIGH rule refused).

## Read requested

This touches `custom_components/omoda9/core/commands.py`, which `CONTRIBUTING.md` puts in
the shared-code row — a comment, but the rule does not have a special case for comments and
I would rather not invent one. The merge is not blocked (R06/R20).

@drake69 — this is the thing you asked for. The part I would most like argued with is the
`R`-to-prose table in `tools/README.md`: several rules turned out to have nothing behind them
in this repository, and I would rather you disagree with how I labelled them than inherit
them by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,10 +13,20 @@ omoda9_diag.off
 # --- runtime / dati locali ---
 data/
 venv/
+.venv/
+.venv-test/
 __pycache__/
 *.pyc
 *.bak
 *.bak_*
+
+# --- tooling: local registers and personal patterns, never tracked ---
+# tools/private-patterns.txt holds the very strings check-secrets.sh exists to catch
+# (a surname inside an e-mail, home paths, machine names): committing it publishes them.
+DEROGATIONS.log
+DEROGHE.log
+COMMENTI.log
+tools/private-patterns.txt
 
 # IDE
 .vscode/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,3 +203,15 @@ If you are working with a coding agent — and most of us are — point it at
 above plus the literal git and GitHub commands for branching, opening a pull
 request, syncing and getting out of trouble. You are not expected to know
 GitHub to contribute here.
+
+Everything above is prose, and prose is a thing to remember rather than a thing that
+stops you. [`tools/`](tools/) is the same rules written so a machine can check them:
+numbered `R01`–`R26`, each one pointing back at the paragraph here or in `AGENTS.md`
+that it comes from, and each one saying whether it blocks, warns, or merely asks.
+`./tools/rules.sh list` prints the table. Where a rule has **no** paragraph behind it in this
+repository, it is marked a house rule of whoever wrote the scripts, rather than
+presented as something the group agreed.
+
+Using it is optional and always will be: `git push` and `gh pr create` are in
+`AGENTS.md` for a reason. The point of putting it here is that a rule only one machine
+can enforce is a rule the rest of us are guessing at.
--- a/custom_components/omoda9/core/commands.py
+++ b/custom_components/omoda9/core/commands.py
@@ -14,8 +14,8 @@ Header: Authorization=<userToken>, timestamp=<ms>, Content-Type=application/json
 
 ⚠️  Ogni send() col taskId valido ATTUA sull'auto. È pensato per essere invocato SOLO
     dal tap di Rino su un pulsante in Home Assistant (= suo consenso esplicito).
-    Catalogo body ricostruito 1:1 dagli envelope reali in
-    /root/omoda9_capture_20260620/command_envelopes.txt.
+    Catalogo body ricostruito 1:1 dagli envelope reali catturati dall'app ufficiale
+    (materiale di cattura: canale privato, mai in questo repo).
 """
 import os
 import json
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -180,7 +180,7 @@ REALTIME_ASLEEP = {"code": "A07900"}
 # Sono omesse le 98 voci FIGLIE del ramo 1 (stato veicolo, `parentId` 101): verificate tutte a 1,
 # e nessun comando le tocca. Restano le altre 182 — i comandi dei rami 2/3/4 più le cinque radici,
 # fra cui il 101 stesso — che sono quelle che decidono.
-# Fonte grezza: /root/omoda9-project/45_capability/permessi_jaecoo7_phev_eu.json
+# Fonte grezza: la lista permessi letta su un Jaecoo 7 PHEV EU (canale privato, mai in questo repo).
 JAECOO7_PHEV_EU = {
     1: 1, 2: 1, 3: 1, 4: 1, 101: 1, 201: 0, 202: 1, 203: 1,
     204: 1, 205: 1, 206: 1, 207: 1, 208: 1, 209: 1, 210: 0, 211: 1,
--- a/tests/test_repo_hygiene.py
+++ b/tests/test_repo_hygiene.py
@@ -4,9 +4,13 @@ Il repo è **pubblico**. Nella stessa cartella di lavoro convivono due mondi:
 
   * il **package pubblicabile** (`custom_components/omoda9/`, README, LICENSE, hacs.json,
     CHANGELOG) — ciò che finisce su GitHub e da lì in HACS;
-  * il **tooling interno** (script di release, handoff, il ponte legacy, questa suite) e
-    i **dati per-account** (certificati, token, `.env`, file del monitor diagnostico) —
-    che devono restare su disco ma fuori dal set tracciato.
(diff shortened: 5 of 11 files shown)


drake69:
**Approved, and it fixes something nobody asked it to.**

**Two absolute private paths leave the tracked set.** `core/commands.py` carried `/root/omoda9_capture_20260620/command_envelopes.txt` and `tests/fixtures.py` carried `/root/omoda9-project/45_capability/permessi_jaecoo7_phev_eu.json`. Neither is a secret on its own, but both describe a private machine's layout in a public repository, and both are now pointers to the private channel instead. That is #7's rule applied retroactively, in a pull request that is nominally about tooling.

**The `isdir` → `exists` change is a real bug fix with a real symptom.** Inside a git worktree `.git` is a file, so the fixture skipped silently — tests green locally, and the hygiene suite simply not running. A skip that looks like a pass is the worst failure mode a guard test has.

**Splitting `VIETATI_ESATTI` into root-only names is correct rather than convenient.** Banning `release.sh` at any depth would have rejected `tools/release.sh` — that is, this pull request would have failed its own rule — and the split keeps the ban where the risk actually is: stale private copies at the root. The comment saying the test discovered the change is the honest version of that story.

**`tools/private-patterns.txt` in `.gitignore` is the important line.** That file holds the very strings `check-secrets.sh` exists to catch; committing it would publish them. Shipping the `.example` and ignoring the real one is the right split.

**No version bump here**, so unlike #32, #25 and #26 this one does not contend for `1.14.0` and can land independently of that ordering.

I read the Python and the diff to `.gitignore`, `CONTRIBUTING.md` and the hygiene test closely; the ~1700 lines of shell in `tools/` I read for shape and for anything that writes outside the repo, not line by line. Saying so because "approved" should carry how far the reading went.

*Per the convention: the review was carried out by the agent I drive. The decision is mine.*

drake69 (approved):
Approved — findings in the review comment above.

PR #38, opened 2026-08-25
docs+test: close issues on merge, and never leave a language behind

drake69:
Two rules for `AGENTS.md`, and the second arrives with the check that enforces it rather than as a request to remember.

## `Closes #N`

**Correcting my own evidence for this, having checked instead of assumed.** I opened this saying that fifteen issues were open and one closed, implying the others lacked the line. They do not: #14, #16, #17, #21 and #33 all carry `Closes`/`Fixes`, and @Caslinovich used `Relates to #1` on #25 and #26 — which is precisely the distinction this rule asks for, applied before anyone wrote it down. The issues are open because those pull requests have not merged yet, not because anybody forgot.

So this rule is not a correction of current practice. It is writing down what four of us already do, so that it survives the fifth person and the tired evening.

The rule says to use it only when the merge really finishes the issue. A partial fix gets `Relates to #N` and a sentence about what is left, so the issue outlives the merge instead of being closed on a half-answer. @Caslinovich, #25 and #26 are exactly that case: both relate to #1 and neither finishes it, since rear defrost (`232`) and the seat category (`214`) are still denied on that car.

## Never leave a language behind

A label or an error message lives in **three** files: `strings.json`, `translations/en.json`, `translations/it.json`. A key present in one and missing in another produces an entity with **no name at all** — but only for the people using that language, so whoever wrote the change never sees it.

Not hypothetical. The fork line carries `en.json` and no `it.json`; porting the charging-energy counters from it without looking would have given every Italian user three unnamed entities.

`tests/test_traduzioni_complete.py` compares the key sets across the three files. **It starts green** — 185 leaf keys, identical in all three — so it is a ratchet rather than a cleanup, in the sense of #6: a rule that is already true, written down so it cannot quietly stop being true.

It deliberately checks **keys and not text**, and the docstring says so: a `name` left in English inside `it.json` passes. A test that appeared to verify translation quality would get trusted for something it cannot do, which is worse than not having it.

## A third rule, added after the fact — and it is the one that produced the other two

`9ec70b8` adds a section to `AGENTS.md`: **before you write a rule, ask what would enforce it.**

It was missing from this branch and that was the joke at my expense. The reasoning below was sitting in a pull request description, which is exactly where a rule goes to be forgotten. The section names the five failures of 24 August, because the abstraction on its own would not survive contact with a tired evening, and ends with three instructions: name the thing that enforces the rule, ship it in the same change, and check the thing exists before telling somebody to use it.

@Caslinovich — the fifth item on that list is that this file forbids AI co-authorship trailers while your commits carry one, because Claude Code adds it by default. I am naming it as an instance of the pattern rather than as a complaint: it is precisely a rule losing to a tool, and nothing in the published history should be rewritten to fix it — `blame` and `bisect` on that history are why we chose this line as the base.

## Why these two and why now

Both are the same shape as the thing this project kept running into yesterday — a standard that is written down while the thing that makes it hold is missing. Five instances in one day, from the review rule that no pull request could satisfy to the three labels the documents told people to use and that did not exist. So the translation rule ships with its test in the same commit, and the `Closes` rule is a line in a template rather than a habit somebody has to acquire.

*Per the convention: the check that the three files currently agree, and the survey of which issues are closed, were run by the agent I drive. Both rules were asked for by me.*

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,14 @@ someone for whom it worked.
 refactor. If the entity count changes, that is a deliberate, announced change —
 and the test will stop you first.
 
+**Never leave a language behind.** Anything that adds or renames a user-visible label or
+an error message touches **three** files: `strings.json`, `translations/en.json` and
+`translations/it.json`. A key present in one and missing in another produces an entity
+with **no name at all** — but only for the people using that language, so whoever wrote
+the change never sees it. `tests/test_traduzioni_complete.py` compares the key sets and
+fails if they diverge; it checks that you did not forget a file, not that you translated
+anything, so a key is only really done when the text is in the right language.
+
 **No Home Assistant imports in `core/`.** The protocol logic must be exercisable
 without Home Assistant installed. This is what lets the sandbox test a change
 without an instance, and it is the rule the whole suite rests on.
@@ -200,6 +208,15 @@ CI be the gate — that is why CI exists — but say so in the pull request.
 git push -u origin HEAD
 gh pr create --fill
 ```
+
+**Put `Closes #12` in the body when the pull request finishes an issue.** GitHub then
+closes it on merge, and links the two forever. Without it somebody has to remember, and
+nobody does: on 24 August this repository had fifteen issues of which exactly one was
+closed — by the single pull request that carried the line.
+
+Use it only when the merge really finishes the issue. If it fixes part of one, write
+`Relates to #12` and say in the body what is left, so the issue outlives the merge instead
+of being closed on a half-answer.
 
 Then fill in the field-test block in the template. If you did not test on
 hardware, say it there rather than leaving it blank:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,43 @@ adding one, you have found an architecture problem: say so instead of encoding i
 **The suite stays green.** Not "green after a follow-up" — green in the pull
 request that changes the behaviour.
 
+## Before you write a rule, ask what would enforce it
+
+You will be asked to write rules — into this file, into `CONTRIBUTING.md`, into a pull
+request description. **A written rule is free and a tool is not**, so the gap between what
+the documents claim and what the repository actually permits widens on its own, silently,
+until somebody tries to follow the rule and hits a wall.
+
+When that happens it does not look like a rule being broken. It looks like a person not
+bothering.
+
+This project collected five of these in a single day, 24 August 2026:
+
+- `CONTRIBUTING.md` said the review that counts is somebody running the change on their own
+  car — while a pull request produced **nothing anyone could install**;
+- both documents told people to use the labels `beta`, `unverified-hardware` and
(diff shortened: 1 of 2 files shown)


GurliGebis:
@drake69 a suggestion (not 100% sure if this matches this) - would it make sense to have a requirement that all pull requests should accompany an issue.

So before creating a pull request, an issue should be opened, explaining the issue.
Followed by the pull request, that links to the issue.

drake69:
@GurliGebis yes — and I think the reason is sharper than "an issue explains the change".
A coordination point is what we are missing, and the cost of not having one is already
sitting in the pull request list.

Twelve open pull requests right now, from three of us in parallel. What they touch:

| file | open PRs touching it |
|---|---|
| `manifest.json` | #14, #25, #26, #32 |
| `strings.json` + `translations/en.json` + `translations/it.json` | #14, #16, #17, #19, #33 |
| `config_flow.py` | #14, #16, #17, #19 |
| `coordinator.py` | #16, #17, #19, #25 |
| `const.py` | #16, #17, #19, #33 |
| `CHANGELOG.md` | #14, #25, #26, #32 |
| `core/session.py` | #16, #17 |
| `core/omoda_auth.py` | #16, #17 |

`manifest.json` is a version line, and four branches bump it. That is not a risk of
conflict, it is a conflict already written, waiting for the second merge to reveal it. The
three translation files are the same shape with five branches on them. Whoever merges
second pays, and it will not be the person who created the situation.

**What issue-first actually buys, and what it does not.** It does not prevent the overlap —
two people can still open two issues and branch onto the same file. What it buys is that the
overlap becomes **visible before three people have branched**, instead of at the merge. Right
now the only place the collision exists is in our heads, and we each have a different half.

So: yes to the rule. I would add two things to it, in the spirit of your suggestion and of
the section we just wrote into `AGENTS.md` — *before you write a rule, ask what enforces it*:

- **A pull request template with the issue link as a required field**, so the rule lives in
  the form rather than in somebody's memory of having read `CONTRIBUTING.md`.
- **The issue says which files it expects to touch.** One line. It is what turns a list of
  issues into something you can actually check a new branch against, and it is the part that
  would have caught the four `manifest.json` branches on the day the second one opened.

I will be honest that nothing here is enforced by a test — this one is a rule about people,
and the template is the strongest thing available. The `Closes #N` check from this pull
request is the enforceable half, and it already landed.

One thing I do **not** want the rule to cost us: your `#11` comment about projects and a
backlog is the same idea one level up, and I would rather not have both a mandatory issue
*and* a board that has to be kept in step by hand. If you set the project up, the issue is
the card and there is one place to look. Happy to follow your lead on that.

GurliGebis:
@drake69 new issues should be added to the backlog automatically (if not, that should be done by copilot).
If you want, I can do the project manager part of it - I don't mind keeping the backlog and "sprint" up to date. (Would need to be coordinated with you guys, so we agree on what should be in the next release) 🙂

PR #43, opened 2026-08-25
ci(beta): cut pre-releases from master only, because HACS offers the newest one

### LINKED ISSUE #42: The beta pipeline can only test one change at a time, never the whole
The beta pipeline can only cut a build from a **pull request**. There is no way to make the accumulated state of `master` installable — and that is the build that would actually catch the failures this project has.

## The gap

`beta.yml` triggers on the `beta` label, or on `workflow_dispatch` with `pr` as a **required** input. Either way the build is `master-as-it-was-when-that-branch-opened` plus one change. Nobody has ever field-tested the changes standing next to each other.

That matters here more than it would elsewhere, because our failures are joint properties rather than local ones: the entity count, the three translation files that have to carry the same key set, two sensors reaching for the same field, capability handling that only bites on a car nobody here owns. A change can be correct on its own and wrong beside another, and every beta we have cut so far was blind to exactly that.

**Measured this afternoon, right after #32, #35 and #37 went in: all ten remaining open pull requests were behind `master` — between 11 and 33 commits — every one of them with green CI.** #6, #16 and #17 were 33 behind. Being up to date is now enforced, which fixes the CI half. It does not give anybody an installable build of the whole.

## What to change

Make the `pr` input optional. Empty means: cut the beta from `master`.

- **empty** → resolve `master`'s head, skip the fork check (`master` is this repository by definition, and it is reachable only through branch protection), and publish release notes that ask the tester about the aggregate rather than about one change, with a compare link against the last stable so they can see what is in it.
- **a number** → exactly what happens today. Unchanged.

The version guard needs nothing: it already refuses to cut a beta whose manifest version is not ahead of the latest stable, and on `master` that is satisfied by whichever release-bearing pull request merged last.

The working cycle this unlocks: **merge a batch → cut a beta from `master` → somebody drives a real car → next batch.** That is the loop this project's review model assumes and has never been able to run.

## Why now rather than later

Rino is away from early September (see #10). The field-test window for anything needing a real car is short, and the first test of the aggregate should not fall after the car is gone. This is a one-input change; it should not wait for the queue to drain.

## Not in scope

Cutting a beta automatically on every merge to `master`. It should stay a deliberate act by somebody who then goes and asks a person to install it — an automatic pre-release nobody asked for is a notification, not a test.

drake69:
Closes #42.

## What changes

`workflow_dispatch`'s `pr` input becomes optional. Leave it empty and the beta is cut from `master`.

- **empty** → resolve `master`'s head, skip the fork check, and publish notes that ask the tester about the aggregate, with a compare link against the last stable so they can see what is in it. Release title: `1.14.0-beta.N — insieme, da master`.
- **a number** → byte for byte what happens today.

The version guard needed nothing: it already refuses a beta whose manifest version is not ahead of the latest stable, and on `master` that is satisfied by whichever release-bearing pull request merged last.

## The bug I hit on the way, which is worth more than the feature

The release notes were built with a heredoc nested inside a command substitution — `note=$(cat <<EOF … EOF)`. It works until the text changes. When I shortened the first paragraph, `bash -n` started reporting an unterminated quote forty lines further down: it had stopped recognising the heredoc, so it was parsing the body as code and hit the apostrophe in `l'ora`. Nothing in the construct was load-bearing, so the notes now go to a file and the release is created with `--notes-file`.

Recording it because the failure mode is nasty: the syntax error appears in a line nobody touched, and it depends on how many lines of prose sit above it.

## How it was checked, since the workflow itself cannot be run from a branch

- every `run:` block in the file extracted and passed through `bash -n` — the whole file, not just the parts I edited, which is how the heredoc was caught at all;
- the publish step run for real with `gh` stubbed out, in both modes, to read the notes and the release title each one produces. The pull-request mode output is unchanged.

What that does **not** cover: the actual workflow run, the `gh api` calls, and whether the zip is built correctly from a `master` checkout. The first real integration build will be the first proof of those, and it should be cut deliberately with somebody ready to install it.

## Why this is worth doing now rather than after the queue drains

Rino is away from early September (#10). Anything that needs a real car has a short window, and the first field test of the aggregate should not land after the car is gone. It is one optional input.

## A note on language

The file is in Italian and I kept it that way, including the new blocks. `AGENTS.md` in #40 says new code is English and that a block you are editing becomes English as part of the change — but #40 has not merged, and half-converting this one file would leave two languages inside a single `run:` block, which reads worse than either. When #40 lands, this workflow is a good candidate for a conversion of its own.

---

*Per the convention: the analysis of the gap, the `bash -n` sweep and the stubbed dry-run were carried out by the agent I drive. The decision that the aggregate needs to be testable, and that it needs to be testable this week, is mine.*

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -1,134 +1,79 @@
 name: Beta
 
-# Una pre-release INSTALLABILE tagliata da una pull request, per farla provare su
-# un'auto vera prima che il cambiamento entri.
+# An INSTALLABLE pre-release cut from `master`, so the assembled thing can be tried on a
+# real car before it is declared stable.
 #
-# Perché esiste. Il gate di questo progetto non è il merge ma la STABILE, e la
-# review di chi non programma è la prova sul campo sulla propria auto. Perché
-# quella regola sia esigibile, chi ha l'auto deve poter installare la PR — e fino
-# a oggi non poteva: `hacs.json` dichiara `zip_release`, quindi HACS installa solo
-# da una release che porta allegato `omoda9.zip`, e quell'asset veniva costruito e
-# caricato a mano una volta per release. L'unica strada era scaricare un branch e
-# copiare i file in `custom_components/` a mano: esattamente la barriera che tiene
-# fuori le persone la cui review vale di più.
+# Why it exists. The gate of this project is not the merge but the STABLE release, and the
+# review of someone who does not program is the field test on their own car. For that rule
+# to be exigible, the person with the car has to be able to install the code — and until
+# this workflow existed they could not: `hacs.json` declares `zip_release`, so HACS installs
+# only from a release carrying an `omoda9.zip` asset, and that asset was built and uploaded
+# by hand once per release. The only other route was cloning a branch and copying files into
+# `custom_components/` — exactly the barrier that filters out the people whose review is
+# worth most.
 #
-# Come si avvia: si mette l'etichetta `beta` sulla PR. Due tocchi dal browser di un
-# telefono — è deliberato, perché chi deve poterla chiedere non sempre ha un
-# computer sotto mano. `workflow_dispatch` resta come uscita di sicurezza.
+# WHY ONLY FROM `master`, AND NOT FROM A PULL REQUEST. This was tried the other way first,
+# on 24 August 2026: four pre-releases cut from four different branches, numbered beta.1 to
+# beta.4. It does not work, and the reason is in HACS itself
+# (`custom_components/hacs/repositories/base.py`, lines 1103-1114 and 385-388):
 #
-# E si avvia anche SENZA una PR: `workflow_dispatch` col campo vuoto taglia la beta
-# da `master`. Serve a collaudare l'INSIEME, che è la cosa che una beta per-PR non
-# può fare: quella prova «master di quando ho aperto il branch, più una modifica»,
-# mai le modifiche una accanto all'altra. E i guasti di questo progetto vivono lì —
-# il conteggio delle entità, le tre liste di traduzione che devono coincidere, le
-# capability: sono proprietà CONGIUNTE, che nessuna singola PR esercita. Il ciclo
-# che serve è: si mergia un lotto, si taglia una beta da master, la si prova su
-# un'auto vera, si passa al lotto dopo.
+#     releases = await self.get_releases(prerelease=True, returnlimit=30)   # GET /releases
+#     for release in releases:
+#         if release.draft: continue
+#         elif release.prerelease:
+#             if self.data.prerelease is None:
+#                 self.data.prerelease = release.tag_name    # the FIRST one it meets
+#     ...
+#     if self.data.show_beta and self.data.prerelease is not None:
+#         available = self.data.prerelease
 #
-# Chi può avviarla: solo chi ha accesso in scrittura, perché è GitHub a decidere
-# chi può etichettare. I tester non avviano niente: ricevono un link e installano.
+# `GET /releases` returns newest-created first, and HACS never sorts. So HACS offers the
+# most recently PUBLISHED pre-release, not the highest version — the version number does not
+# enter the decision at all. A tester with "show beta versions" on does not choose which
+# build they get: they get the last one published, whatever branch it came from. Cutting
+# betas from branches therefore produces a counter that rises while the content moves
+# sideways: beta.4 contained one pull request and none of the three before it.
+#
+# The consequence, and the rule this file now enforces: every published pre-release must be
+# coherent on its own, and only a build from `master` is. That also happens to be the thing
+# a per-PR beta could never do — test the changes NEXT TO EACH OTHER. The failures of this
+# project live exactly there: the entity count, the three translation lists that must agree,
+# the capabilities. Those are JOINT properties that no single pull request exercises.
+#
+# The cycle this supports: merge a batch, cut a beta from master, try it on a real car, move
+# to the next batch.
+#
+# What was lost with the per-PR build, and is not yet replaced: putting a change written for
+# a car nobody here owns in front of its owner BEFORE it is merged. See the discussion on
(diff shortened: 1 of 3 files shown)


GurliGebis (changes requested):
Sounds good - only think I worry about is if we make beta 5 from master, and beta 6 from a feature branch, and then beta 7 from another branch, we end up with things getting lost (things added in beta 6 would be missing from beta 7).

So I think the best solution is to only make betas from the master branch.
I know this could result in broken code being merged - but if things gets merged into master, they should be tested and fixed.
Worst case, they can be reverted from master, and then recreated in a new branch, fixed, and then merged again.

Based on the size of this project, that shouldn't be too big of an issue.

What are your thoughts on this @drake69 ?

drake69:
@GurliGebis — you are right about the failure mode, and I want to separate it from what this pull
request actually does, because I think we agree more than the review state suggests.

**Your scenario is real.** If a tester keeps installing whatever pre-release is newest, and those
come alternately from `master` and from feature branches, then beta.6 from branch A and beta.7 from
branch B each carry the *other one's* work missing. Version numbers go up while content goes
sideways. That is a genuine trap and it is worth closing.

**But this pull request does not create it — it is purely additive.** Cutting a beta from a pull
request is what `beta.yml` does on `master` today; this only adds the missing second mode, the one
that builds the *aggregate*. Removing the per-PR mode is a separate change to the same file, and
one I would rather decide on its own merits than as a side effect.

**The reason I would not remove it yet.** Per-PR betas exist for a specific case: code written for
a car nobody here owns. Half of what we merge is `unverified-hardware`, and the only way to put it
in front of the person whose car can disprove it *while it can still be changed* is to make the
branch installable. That is the mechanism in `CONTRIBUTING.md`, and "merge it into master first,
revert if broken" swaps a pre-merge check for a post-merge repair on people's real cars.

**Concretely, so the trap you named cannot happen anyway:** the two modes are asked for in different
ways — a `beta` label on a pull request, versus running the workflow with the field left empty — and
they say different things in the release notes. A per-PR build asks "does the car do what this PR
claims?". A master build asks "does everything merged so far hold together?". Nobody is meant to
*track* per-PR builds; they answer one question and are done.

**What I am asking for.** I have merged this so the aggregate mode exists — the whole point is that
we have never once tested the assembled thing, and Rino is away, so the window is now. If you still
want per-PR builds gone after seeing the two side by side, open it as its own pull request and I
will not argue the general point: you are right that a beta line has to be monotonic. I just do not
want the missing half blocked on removing the existing half.

**Disclosure, since it matters for how much the above weighs:** the reading of `beta.yml` and the
conflict analysis behind this were done by my agent. The decision to merge over your review, and
the argument for keeping both modes, are mine.

drake69:
@GurliGebis — you were right, and my comment above was wrong. Here is the measurement that
settles it.

I argued that per-PR and master betas can coexist because "nobody is meant to track per-PR
builds". That claim is false, and I should have checked it before writing it rather than after.

**What HACS actually does** — `custom_components/hacs/repositories/base.py`, lines 1103-1114 and
385-388:

```python
releases = await self.get_releases(prerelease=True, returnlimit=30)   # GET /releases — unsorted
for release in releases:
    if release.draft: continue
    elif release.prerelease:
        if self.data.prerelease is None:
            self.data.prerelease = release.tag_name    # the FIRST one it meets
    else:
        self.data.last_version = release.tag_name
        break
...
if self.data.show_beta and self.data.prerelease is not None:
    available = self.data.prerelease                   # and that is what it offers
```

`GET /releases` returns newest-created first. So **HACS offers the most recently published
pre-release, not the highest version.** The version number never enters the decision —
`AwesomeVersion` is not consulted on this path at all. Anyone can re-run this:

```
gh api repos/chery-connect-ha/omoda9-ha/releases \
  --jq '[.[] | select(.draft==false)] | map(select(.prerelease))[0].tag_name'
```

returns `v1.14.0-beta.4` today.

**This is not hypothetical, it already happened here.** On 24 August we published beta.1 (#14)
at 12:49, beta.2 (#25) at 18:18, beta.3 (#26) at 18:20, beta.4 (#32) at 18:21. Anyone with
"show beta versions" on was moved to beta.4 — master plus #32, containing none of the other
three. @kowi4 reported #48 and #49 from beta.2. The counter went 1 to 4 while the content moved
sideways, which is your sentence: *things added in beta 6 would be missing from beta 7*.

**So the conclusion is yours, and stronger than the argument you gave for it:** if every
published pre-release is pushed to everyone with the flag on, then every published pre-release
has to be coherent on its own — and only a build from `master` is.

**Done, in `682ea47` just pushed here.** This pull request now does both halves: it adds the
master mode and removes the `beta` label trigger. The workflow went from 293 lines to 200 — no
PR resolution, no fork guard, no label listener, one job. `AGENTS.md` and `CONTRIBUTING.md` are
updated in the same commit, because the rule in `AGENTS.md` is that a change contradicting a
document changes the document with it.

**What that costs, and what I would like your help with.** The label existed for one case: code
written for a car nobody here owns — roughly half of what we merge is `unverified-hardware`. It
put the change in front of the one person whose car can disprove it *while it can still be
changed*. Merging first and reverting if broken moves that repair onto people's real cars, which
is worse for exactly the contributors we can least afford to lose. So I would like to keep a
pre-merge field test, just not as a published pre-release of this repository. You know Actions
better than I do — is there a shape that works? An artifact plus install instructions is the
obvious one, but it reintroduces the manual-copy barrier that filters out precisely the testers
who have the car. I have written it into both documents as an open problem rather than pretending
the trade is settled.

Your review is still the blocking one on this PR, which is correct — the change you asked for is
now in it. Over to you.

GurliGebis (approved):

GurliGebis:
If it turns out to be a problem, a way to do it would be to have a test branch, that is reset every time a new release is made, where feature branches can be merged into, and beta releases be made from.
Then, when features work, they can be merged to master.

It requires quiet a lot of maintenance (run several projects at work like this - it requires resetting test to master from time to time, and merging not-yet-done stuff into it again).

Lets see what happens with this 🙂

Issue #45, opened 2026-08-25
Half the threads here are unreadable to some of the people reviewing them


drake69:
`AGENTS.md` says the project is already English and names commit messages, pull requests and the documents. It does not name **issues, issue titles, comments and reviews** — and those are where most of the conversation actually happens.

The result is a repository half of which some of its own maintainers cannot read.

## The evidence, from the current state of this repo

Open issues and pull requests with Italian titles or bodies right now: #25, #26, #5, #2. Meanwhile @GurliGebis, @JackRonan, @ThomasMeyer1970, @Sisku and @kowi4 write in English — and @GurliGebis, who is building the merged repository and offered to do the HACS default-store submission, has no Italian at all.

So a maintainer opens the queue and finds threads he cannot read, on code he is being asked to review. Nobody decided that; it accumulated because nothing was written down.

## The rule to write down

> Everything addressed to the people working on this project is in English: issues and their titles, pull requests, comments, review comments, commit messages.

**And the exception, which matters as much**, because otherwise somebody will helpfully "fix" it: the Italian written **for users** stays Italian. `translations/it.json`, the Italian half of `CHANGELOG.md`, and command-result text are translations doing their job, not a lapse. The test is who the sentence is addressed to — a collaborator, or somebody driving the car.

## What enforces it

A reader, and `AGENTS.md` already says so for the equivalent rule about code: no check can tell whether a sentence is English. The one automated piece that exists is the hook refusing non-English commit messages. This rule inherits both the honesty and the limitation, and should be written as such rather than pretending a test could hold it up.

GurliGebis:
Thank you 🙂
So far I've been able to follow along somewhat using Google Translate, but I think other people will just give up without trying.

So I think going with English as the primary coding and communication language would be a good choice. (I agree on not changing everything in code, but doing it when things gets touched, so it slowly transitions to English).

One more thing - I think I was it mentioned somewhere - but should we make sure everything visible to the user is translated to English (naming of the things exposed in the Home Assistant UI) before we change the domain? (I'm pretty sure we cannot change it after the fact, without breaking it for everyone)

PR #46, opened 2026-08-25
docs+ci: one pull request raises the version, and a check that says so

### LINKED ISSUE #44: Nothing says who bumps the version, so everybody did
Nothing in `AGENTS.md` or `CONTRIBUTING.md` says who raises the version in `manifest.json`, or when. Two mechanical constraints exist and neither answers the question: `test_versione_manifest_e_changelog_allineati` requires the manifest version to have a `CHANGELOG.md` section, and `beta.yml` refuses to cut a pre-release whose manifest is not ahead of the latest stable.

So four people reached four different answers at once, and it cost a working afternoon.

## What happened, since it is better evidence than the argument

On 25 August, #32, #25 and #26 each bumped `manifest.json` to `1.14.0` and each opened their own `## v1.14.0 — 2026-08-24` section. Only one of them could have that number. #32 merged; #25 and #26 went from `MERGEABLE` to `CONFLICTING` in the same minute. #14 and #19 were carrying the same collision behind them. Five pull requests, one line, three authors — every one of whom had to be asked to redo work that was already correct.

It is also a question no author can answer alone: whether your change is a patch or a minor depends on what else ships beside it, which you do not know when you open the branch.

## The rule to write down

> `manifest.json` carries the number of the **next** release, and exactly one pull request ever bumps it: the first one opened after a stable ships. Everybody else adds their entry to the existing `## v<next>` section and leaves the version alone.

**On the wider practice**, so this is a deliberate adaptation and not an invention: most mature Home Assistant custom integrations bump at release time — either in a dedicated release pull request, or by keeping a placeholder in `manifest.json` and injecting the real number from the tag at build time. Per-pull-request bumping is uncommon and is generally regarded as a mistake, for exactly the reason above.

We cannot adopt the placeholder form as-is, and the reason is one we chose on purpose: `beta.yml`'s guard compares the manifest against the latest stable, because in semver `v1.13.0-beta.1` sorts *before* `v1.13.0` and HACS would present it as a downgrade. That guard requires a real number in the file, ahead of the stable, before any beta can be cut. So the rule above is the minimal correct adaptation — one bump per release, as everyone else does, just early enough for our own pipeline.

Worth noting that `beta.yml` already injects the version into the zip it builds, so half the machinery for the canonical form is there. Moving the rest — placeholder in the repo, injection from the tag for stables too, and a guard that compares tags instead of the manifest — is coherent and is **not** proposed here.

## What this does not fix

`CHANGELOG.md` stays shared: everyone writes into the same section and those edits still collide. That is an ordinary additive conflict rather than the same line carrying three different values, and it is worth being honest that the rule reduces the problem instead of removing it.

drake69:
Closes #44.

Two files: the rule in `AGENTS.md`, and the check that holds it up.

## The rule

> `manifest.json` carries the number of the **next** release, and exactly one pull request ever raises it: the first one opened after a stable ships. Everybody else adds their entry to the existing `## v<next>` section and leaves the version alone.

Today that number is `1.14.0`, taken by #32. So #14, #19, #25 and #26 fold into it and drop their second bump.

**What it fixes:** `manifest.json` stops being a contended file — one bump per release instead of one per pull request. That was one of the two files that made five parallel branches expensive this week.

**What it does not fix, and the text says so:** `CHANGELOG.md` stays shared. Everyone writes into the same section and those edits still collide — but as an ordinary additive conflict, not one line carrying three different values.

## The check

`AGENTS.md` says a rule without something enforcing it is a hope, so this one arrives with `version-guard.yml`. On a pull request touching the manifest it compares the new version against the base, and:

| situation | result |
|---|---|
| version unchanged | passes, silently |
| base was still on the last stable | passes — this is the legitimate first bump |
| the next number is already set | **fails**, naming the number to restore and the section to write into |
| no stable released yet | passes |

It deliberately does **not** judge which number was chosen. When a bump is due, any number passes: which one it should be is a release decision, not a workflow's.

Verified against all four cases with the `git` and `gh` calls stubbed, since the workflow cannot be run from a branch. What that does not cover is the real `gh api` call and the `paths:` trigger — the first pull request that touches the manifest will be the proof of those, and given the rule, that is #14.

## Two choices worth flagging

**This file is in English while `beta.yml` and its neighbours are Italian.** Deliberate: the error message is read by whoever trips over it, and not all of us read Italian. #45 proposes making that general; this file does not wait for it.

**On the wider practice**, so the rule reads as an adaptation rather than an invention: most mature Home Assistant custom integrations bump at release time, either in a dedicated release pull request or by keeping a placeholder in the manifest and injecting the number from the tag at build time. Per-pull-request bumping is uncommon and generally considered a mistake. We cannot take the placeholder form as-is, because `beta.yml`'s guard needs a real number ahead of the stable before any beta can be cut — a constraint we chose on purpose and documented at length. Hence one bump per release, as everyone else does, only early enough for our own pipeline. Worth knowing that `beta.yml` already injects the version into the zip, so half the machinery for the canonical form exists; moving the rest is coherent and is **not** proposed here.

---

*Per the convention: the collision was measured by the agent I drive across the open pull requests, and it wrote and exercised the guard. Deciding that this needed a rule rather than four separate answers is mine.*

### DIFF
--- a/.github/workflows/version-guard.yml
+++ b/.github/workflows/version-guard.yml
@@ -1,0 +1,71 @@
+name: Version guard
+
+# Exactly one pull request raises `manifest.json`: the first one opened after a stable
+# ships. Everybody else adds their entry to the section already open and leaves the
+# version alone. The rule lives in AGENTS.md; this workflow is the thing that holds it
+# up, because a rule without one is a hope (AGENTS.md, "Before you write a rule").
+#
+# Why it exists, in one line: on 25 August 2026 three branches each declared `1.14.0`
+# and each opened their own changelog section. The first merged; the other two went
+# from mergeable to conflicting in the same minute, and two more were carrying the same
+# collision behind them. Five pull requests, one line, three authors.
+#
+# This file is in English, unlike its neighbours, and deliberately: the message below is
+# read by whoever trips over it, and not all of us read Italian.
+#
+# What it does NOT do: judge whether the number chosen is the right one. When the version
+# is due to be raised — that is, when the manifest is still sitting on the last stable —
+# any number passes, because which one it should be is a release decision, not a
+# workflow's.
+
+on:
+  pull_request:
+    paths:
+      - "custom_components/omoda9/manifest.json"
+
+permissions:
+  contents: read
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Only one pull request raises the version
+        env:
+          GH_TOKEN: ${{ github.token }}
+          BASE: ${{ github.event.pull_request.base.sha }}
+        run: |
+          set -euo pipefail
+          MANIFEST=custom_components/omoda9/manifest.json
+
+          new=$(jq -r .version "$MANIFEST")
+          base=$(git show "$BASE:$MANIFEST" | jq -r .version)
+
+          if [ "$new" = "$base" ]; then
+            echo "Version unchanged ($base): nothing to check."
+            exit 0
+          fi
+
+          # The version DID change. That is legitimate only if the number had not been
+          # raised yet since the last stable — that is, if the base manifest was still
+          # sitting on it. `releases/latest` excludes pre-releases, which is what we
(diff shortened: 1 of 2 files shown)


GurliGebis (approved):
LGTM

Issue #41, opened 2026-08-25
Remove the bundled mutual-TLS client certificates from this repository


drake69:
`custom_components/omoda9/certs/store.json` ships mutual-TLS material for 40 regional
brokers: `ca.pem`, `client.pem` and `client.key` for each. One of those three is a
**private key that belongs to Chery**, obfuscated with a fixed keystream that we ship in
the same file.

`SECURITY.md` currently says this about it:

> **The bundled certificates and signing constants are not secrets of yours.** The
> mutual-TLS certificates and the app's signing constants are extracted from the publicly
> distributed vehicle app and are identical for every user. They are not per-account
> material, and finding them in this repository is not a vulnerability.

**Every sentence there is true, and the conclusion is still wrong.** It is true for *your*
security: I measured it, and account isolation is entirely in the MQTT username, password
and topic ACL (`coordinator.py:721-731`) — holding the shared certificate gets nobody into
anybody else's messages. What that paragraph does not say is that the file is **not ours to
publish**. It is a third party's private key, redistributed by us, in public, under our
name. That is not a CVE. It is the kind of thing that gets a repository taken down rather
than patched, and it is entirely within Chery's power to make it our problem at any moment.

So I propose changing the last line of that paragraph from *"finding them in this
repository is not a vulnerability"* to **"we are removing them"**, and this issue is the
plan for doing it.

## Before proposing anything I checked whether we could just delete it

The hypothesis worth falsifying was that the bundle is only *bootstrap* — that somewhere in
the `getTuserId → loginTSP` chain the app trades it for a per-device identity, in which case
every user could get their own and the file goes with nothing lost. Four probes, all
re-runnable, all read-only, none of them touching the car:

**1. What the certificate is.** All 40 regions carry the same subject —
`CN=client, O=EMQX, L=ShangHai, ST=SH, C=CN` — with no SAN, no key usage, no extensions,
and 3650 days of validity each. 40 distinct serials, 40 distinct keys, 40 distinct issuing
CAs. **One identity per region.** Nothing per-device, per-account or per-VIN anywhere in it.

**2. Whether the broker actually wants it — and this is the one that changes the answer.**
On port 8083 the TLS 1.3 handshake **completes even with no client certificate**. If I had
stopped there I would have opened this issue proposing we simply delete the file, and I
would have been wrong. Under TLS 1.3 the client finishes its side before the server has
judged what it received, so what separates acceptance from rejection happens *after* the
[...]

PR #47, opened 2026-08-25
tools+docs: four probes, and what they say about the bundled certificate

drake69:
Relates to #41 — the evidence behind it, so the conclusion can be disagreed with rather than taken on trust.

Nothing in `custom_components/` changes. The scripts live in `tools/` so they never reach the HACS payload, which is built from the package directory alone.

## What is here

| script | question | network |
|---|---|---|
| `step1_cert_inspect.py` | per-device identity, or per-region constant? | none |
| `step2_tls_probe.py` | does the broker demand a client certificate, and does it check who signed it? | TLS handshake only |
| `step3_response_scan.py` | do the provisioning responses carry credential material? | read-only BFF calls |
| `step4_keygen_scan.py` | does anything generate a key pair or a CSR? | none |

Plus `docs/enrollment-findings.md`: the reasoning, the numbers, and the limits.

## The one thing to read carefully, if you read one thing

**Step 2 nearly produced the opposite answer.** On port 8083 the TLS 1.3 handshake completes with **no client certificate at all**. Reading only that gives "the certificate is irrelevant, remove the bundle" — the first row of the decision table in #41 — and it is wrong.

Under TLS 1.3 the client finishes its side before the server has judged what it received, so the verdict arrives afterwards. Three attempts per variant, all consistent:

| what we present | what the broker does |
|---|---|
| nothing | handshake OK, **connection closed within milliseconds** |
| the bundled certificate | handshake OK, **connection held open** |
| a self-signed certificate made on the spot | handshake OK, **connection closed** |

EMQX with `verify_peer` and `fail_if_no_peer_cert`. It never sends alert 116, which is why the probe waits and reads instead of trusting the handshake result. And the self-signed row is the one that closes the question: the broker does not want *a* certificate, it wants one chaining to the region CA.

## Safety rules the scripts obey

No command, no `taskId`, no PIN, no OTP — nothing reaches the car. `step3` reuses the stored session and does **not** refresh it by default: it copies the token file so a running instance never has its refresh token rotated underneath it, and an expired token stops the probe rather than prompting. Values are reported by shape — type, length, an eight-character prefix — and the prefix is suppressed for anything carrying an account identifier. The bundle is de-obfuscated in memory; when OpenSSL needs a key on disk it goes to a private temporary directory and is overwritten and removed.

## Limits, stated in the report rather than implied

One region probed on the network — the EU broker this account uses; step 1 covers all 40. Ports 8883, 8884 and 1883 timed out at TCP level, and a timeout is not a closed port, so nothing is concluded from them. `loginTSP` could not be run at all: there is no HTTP call to replay, since it is native and a no-op in the DEX. That is reported as a result rather than filled in by inference — an invented result would be worse than a missing one, given a decision depends on this.

Two things surfaced that belong elsewhere and are recorded, not acted on: `tls_insecure_set(True)` in `coordinator.py` is no longer needed (full verification against the bundled CA with `check_hostname` on succeeds today), and `certs_dir` is derived from `DOMAIN`, so removing the bundle and renaming the domain must never share a release.

---

*Per the convention: the probes were written and run by the agent I drive, including the control that overturned its own first reading. Deciding to test the hypothesis rather than assume it, and what to do about the answer, is mine.*

### DIFF
--- a/docs/enrollment-findings.md
+++ b/docs/enrollment-findings.md
@@ -1,0 +1,270 @@
+# Does the MQTT stack support a per-device enrollment?
+
+**Measured on 2026-08-25**, against the EU region
+(`tspemqx-app-eu.cheryinternational.com`), with the probes in
+[`tools/enrollment_probe/`](../tools/enrollment_probe/). Every number below can be
+re-produced by re-running them; where a step could not be run, it says so instead
+of guessing.
+
+## Why this was asked
+
+`custom_components/omoda9/certs/store.json` carries mutual-TLS material — `ca.pem`,
+`client.pem`, `client.key` — for 40 regional brokers. It comes from the public
+assets of the official APK, and it is the same for every user of a region. That
+means this repository redistributes a **vendor private key**. The goal is to stop
+doing that.
+
+The hypothesis worth falsifying was that this material is only *bootstrap*: that
+somewhere in the `getTuserId → loginTSP` chain the app trades it for a per-device
+identity — a locally generated key plus a CSR, or a dynamic registration. If such
+an enrollment exists, every user can obtain their own material and the bundle can
+go without losing anything.
+
+## Verdict
+
+**The enrollment does not exist, and the certificate cannot simply be dropped
+either.** Both halves matter, and the second one contradicts the outcome the
+investigation was hoping for:
+
+* the bundled certificate is a per-region constant, not a per-device identity
+  (step 1);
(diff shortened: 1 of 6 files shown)


GurliGebis:
Just got an update to the android app, where it mentions updated certificates.
Wonder if it is due to this? 

Discussion about it created in the reverse engineering repo.

PR #50, opened 2026-08-31
Add Dutch translation

wimb0:
## Add Dutch translation

On my install. some sensors were in English, some in Italian. I added Dutch translations.
I used AI to translate IT and EN to NL, and manually checked which translation was the best.
Also did manual fixes.

## Evidence

Fill in the row that applies and delete the others. Being honest here costs
nothing; overstating it is the only mistake in this project that reaches a
stranger's car.

- [X ] **Verified on hardware** — I own this car and I watched it act.
      Car / region / version: Jaecoo 7 SHS-P  in NL

# Checks

- [ ] Suite green (`pytest tests/ -q`, or CI if the local Python is older than 3.13)
- [ ] Entity count unchanged, or the change is deliberate and named above
- [ ] No VIN, token, certificate, account id or raw capture anywhere in the diff
- [ ] Fails open: nothing here can take away a function that worked for somebody
- [ ] Design docs updated if this contradicts them

PR #58, opened 2026-09-06
Recover the Dutch translation from #50, with its author intact

drake69:
**Recovery of #50, which was destroyed by a fork sync this morning.** Both commits are
@wimb0's, unchanged, with authorship preserved:

```
ac9244c wimb0  Add Dutch translations (nl.json)
be16f93 wimb0  Update phone country codes in Dutch translations
```

### What happened

At 08:52 today the head branch of #50 was force-pushed to this repository's `master` and
the pull request closed in the same second. The cause is not a mystery: the Dutch commits
sat on the `master` of the fork, so syncing the fork could not fast-forward and discarded
them.

**That was our doing.** My comment on #50 said the branch would need `master` brought into
it before it could land. From a browser, on your own `master`, the only thing that
sentence can mean is the Sync fork button, which offers "Discard commits" in exactly this
situation. The goal was right and the instruction was missing.

It is also, precisely, the failure mode #55 was opened to document, happening the day
after, to the contributor whose pull request prompted writing it down. The document was
right and it arrived one day late.

### This pull request is a placeholder for its author

@wimb0 should own this if he wants it. It is his work, and I would rather hand it back
than carry it on his behalf. He can push to this branch, or rebuild it in his own fork
from `99e355d` and open a fresh pull request, in which case this one closes.

### Not ready to merge yet, and not because of anything he did

`nl.json` is 25 keys behind `strings.json`, all of them added by #16 and #19 which merged
after this branch was written. That becomes 30 once #56 lands, which fixes four labels we
were missing ourselves.

A proposed Dutch text for all of them is in a comment on #50: 9 of the strings are his own
words reused, the rest are drafts written without a Dutch speaker and need his eye before
anyone merges them.

### And one thing this makes obvious

The completeness test does not look at `nl.json` at all: its file list is hardcoded to
`strings.json`, `en.json` and `it.json`. A fourth language is unguarded from the day it
lands. That is being tracked separately, but it should be fixed before this merges rather
than after.

### DIFF
--- a/custom_components/omoda9/translations/nl.json
+++ b/custom_components/omoda9/translations/nl.json
@@ -1,0 +1,512 @@
+{
+  "config": {
+    "step": {
+      "user": {
+        "title": "Omoda 9 / Jaecoo",
+        "description": "Kies hoe u zich aanmeldt bij uw Omoda/Jaecoo-account. Gebruik de telefoonoptie als uw account is geregistreerd met een telefoonnummer in plaats van een e-mail.",
+        "menu_options": {
+          "login_email": "Aanmelden met e-mail",
+          "login_phone": "Aanmelden met telefoonnummer (SMS)"
+        }
+      },
+      "login_email": {
+        "title": "Aanmelden met e-mail",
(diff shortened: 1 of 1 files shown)


GurliGebis:
LGTM - just a language json file, no risk in just merging.

Issue #54, opened 2026-09-04
MQTT reports `car_connected: true` but receives zero messages for weeks (REST probe/poll works fine in parallel)


michelangelomonako-cmyk:
Integration version: v1.13.0 (also reproduced on v1.5.24 → v1.12.1 → v1.13.0, same symptom throughout)
Home Assistant: 2026.7.4, Home Assistant OS, Docker/amd64
Summary
The MQTT client connects successfully (`car_connected: true`) but never receives a single message on the subscribed topic. `last_seen` stays frozen indefinitely (currently stuck at a timestamp from over a month ago), while `fields_count` is permanently `0`. Meanwhile the REST probe/poll channel (`async_probe`, telemetry poll every 5 min) works perfectly and keeps odometer, battery, charging status, GPS etc. fully up to date.
Evidence (from Download Diagnostics)
```
"token_present": true,
"certs_present": {
  "ca.pem": true,
  "client.pem": true,
  "client.key": true,
  "eu_prd_cheryinternational.cer": false
},
"state": {





  "session_ok": true,
  "session_detail": "Sessione attiva ✅",
  "awake": false,
  "car_connected": true,
  "has_position_fix": true,
  "last_seen": null,
  ...
  "probe_status": "🟢🛰️ SVOLTA: dati realtime ricevuti ad auto sveglia! ...",
  "fields_count": 0,
  "fields": {}
}
```
`certs_present`: all 3 required mutual-TLS certs present and valid (`ca.pem`, `client.pem`, `client.key`).
`session_ok: true`, PIN/commands work correctly (confirmed via successful lock/unlock commands during troubleshooting).
`car_connected: true` at the MQTT client level (`on_connect` succeeds).
`last_seen` never updates — no `_on_car_message` callback ever fires.
REST-based `realtime` data (from `async_probe`) is fully populated and fresh on every poll cycle, confirming the car itself, the account, and the cloud backend are all reachable and functioning.
Timeline / what has been tried
Initial symptom noticed mid-August 2026 (integration on v1.5.24 at the time).
Updated integration v1.5.24 → v1.12.1 (later → v1.13.0). No change.
Multiple full Home Assistant restarts (not just integration reload). No change.
Re-authentication (new OTP) after a session expiry (triggered by opening the official app once, as documented). Session recovered fully (REST/session_ok back to true), MQTT stayed disconnected/silent.
[...]

PR #66, opened 2026-09-14
Restore realtime-to-fields merge so doors/lock show on a parked car

illusive-c:
## What changes, and why

Doors, windows, lock, engine and plug read from coordinator fields, which a parked car never fills via MQTT. The realtime probe already carries those fields with the same 0/1 encoding but _on_probe_data drops them, so they sit at unknown while battery/range/speed (read from realtime directly) work fine.

This restores the v1.5.28 behaviour of the retired line: merge realtime state keys into fields on every probe, and always run the 15s startup seed even with automatic update off (read-only, no command).

Relates to #54 (that issue is the same empty-fields symptom; this does not fix a silent MQTT channel itself, it makes the states visible again via realtime).

## Evidence

- [x] **Verified on hardware** — Tested on Jaecoo 7 in the UK. Car lock state is show in HA.
- [ ] **Backend only** — n/a, read-only change, no command sent.
- [x] **Not testable by me** — written from code reading; needs field test on an Omoda 9 / Jaecoo. Maintainer: please add the beta label so it can go on a real car.

## Checks

- [ ] Suite green — local sandbox has no pip/pytest, leaving this to CI (local Python 3.13.5, py_compile passes)
- [x] Entity count unchanged, no entity added or removed
- [x] No VIN, token, certificate, account id or raw capture in the diff
- [x] Fails open: only ever adds field values from realtime, never removes MQTT values; absent fields still fall back to restored/unknown
- [x] No user-visible label renamed, no translation change needed

The code reading behind this (coordinator _on_probe_data vs the retired line) was done by the agent I drive. The symptom observation and the decision to fix it this way are the reporter's.


illusive-c:
Field test report from the reporter: tested on their own Jaecoo 7 (UK) with the patched coordinator (same logic as this branch). Doors/lock/windows went from unknown to real states, and a live change on the car was followed in HA. Read-only change, no command sent. The reading/analysis behind the fix was done by the agent; the test and conclusion are the reporter's.

